Issue #1, opened 2022-10-29
Translations using gettext PO


securearth:
Translations can be added using gettext module so people can contribute translations in their native language

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2022.11.0-beta2

bordam:
@nlogozzo @zothma We should have two separate strings for "New": one for when it's referred to Group and one for Transaction since in some languages you translate them differently.

bordam:
Also, for Group I would use a separate string too: in the shortcuts dialog it's better to use a plural form in my language.

bordam:
Missing localization in string "Money Account (*.nmoney)" in open account dialog

bordam:
Like Group, Income and outcome need two separate strings in my language: one for the Type in the transaction dialog and one for Overview.

securearth:
> Also, for Group I would use a separate string too: in the shortcuts dialog it's better to use a plural form in my language.

In some languages, both plural and singular form of group have the same word, like in Hindi

bordam:
@securearth Yeah of course, you can still translate the strings the same way.

securearth:
I just tested my Hindi translation and found out that you have hardcoded "Money", and added translation for "About", instead of using the string "About Money". In English, about is used before the app name, but in languages like Hindi, "बारे" is used after the app name, so it would be "Money बारे" or "Money के बारे में", but using the current schema, it is rendered as "बारे Money" and not "Money बारे". I think you should instead use the string "About Money" instead of hardcoding 'Money', you can add a comment there asking to not translate the word 'Money' and just keep it like "Money बारे" (generally whoever learns to translate, is taught on articles on Internet that as a convention you should not translate the app name).

nlogozzo:
FYI Please submit a PR with your translations by Nov 10, 2022 to be included in the V2022.11.0 stable release for this week.

nlogozzo:
@zothma Is also currently fixing the translation context with the changes mentioned here. Once that gets merged, I will be asking current PRs to update their translations to the new context before merging them :)

nlogozzo:
Thank you everyone for your help with translating this application, making it accessible to more people than ever before. A stable release will be released this week by Nov 10, 2022 with your translations :)
I'm closing this issue as completed since the translation infrastructure is complete.

Issue #2, opened 2022-10-30
Implement Groups and Envelope Budgeting


nlogozzo:

ovari:
Does this issue refer to `Categories`? If so, will subcategories and nested categories be implemented? Thanks

nlogozzo:
Yes, it refers to categories as well.
What I'm thinking of doing is implementing a "Groups" section on the AccountView. In this section you can add, edit and delete groups. Groups will have a name, color, and optional monthly balance:
- If users just want to group they're transactions they will create a group with a name, color, and unlimited monthly balance.
- If users would like to use envelope budgeting, they can add a monthly balance to the group
Then from the Transaction POV, the user can select a group when adding/editing a transaction.

nlogozzo:
As far as subcategories and nested categories, I don't know if that will be implemented.
When I think of groups, I see people creating groups such as Travel, Holidays, Groceries, Bills, etc...
If you can provide me with some use cases for subgroups/nested groups I will consider it.

nlogozzo:
![image](https://user-images.githubusercontent.com/17648453/199515661-7d6914d8-0fe2-44d8-9107-be3b96557695.png)
![image](https://user-images.githubusercontent.com/17648453/199515721-be4ff93b-9fa9-4713-a1fb-4cd6005d2816.png)

Some progress: adding/editing a group
Just need to allow associating a transaction with a group.

ovari:
> If you can provide me with some use cases for subgroups/nested groups I will consider it.

https://github.com/moneymanagerex/moneymanagerex/issues/1477

nlogozzo:
![image](https://user-images.githubusercontent.com/17648453/200148405-f24dedd4-083d-4a27-859a-e159c75d9e65.png)

nlogozzo:
![image](https://user-images.githubusercontent.com/17648453/200148407-e6511e4a-2ebd-4891-8c01-c50132019aa2.png)

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2022.11.0-beta1

ovari:
Perhaps your `Group` field is equivalent to the `Class` field in Quicken. The `Class` field can not be nested.

![Quicken Class](https://user-images.githubusercontent.com/17465872/200154766-2658cc76-0dd6-411d-8dbd-29dd4846f86c.png)

[...]

PR #9, opened 2022-11-05
Translation using gettext

zothma:
# What?
Added support for internationalization using GNU gettext and .po files, as suggested in #1 

# How?
This version adds support for gettext directly within meson, and allows to compile the following useful targets :
 - `org.nickvision.money-pot`: creates a .pot file, a model used by translators to create their .po files
 - `org.nickvision.money-update-po`: updates existing .po files by automatically adding new entries to be translated while keeping the other entries safe
All the other targets can be shown by running `meson introspect --targets` and compilation can be done by running `meson compile <target>`

Inside the code, each string containing displayed text is now wrapped with `_(...)` macro. Whenever the string was too large (like the XML string for shortcuts) or was split using concatenation, I used a new formatting function which acts like sprintf in C but converts back to native C++ strings.

Finally, each file using gettext is manually added to the po/POTFILES to ensure that .po files are updated correctly.

A lot of files have been modified, but it is mainly because of wrapping strings with the gettext function and adding the proper header. Here are the files that deserves to be looked :
 - **src/utilities/string_format.hpp**: new string formatting function
 - **src/utilities/translation.hpp**: header file for the gettext function
 - **src/ui/views/shortcutsdialog.cpp**, **src/controllers/accountviewcontroller.cpp** and **src/ui/views/transactiondialog.cpp**: implementing at some point `string_format`
 - **meson.build** and **po/meson.build**: i18n dependency addition

# What to do next?
This version doesn't support for international date format (dd/mm/yyyy for instance) and does not allow to translate the about page, as it is generated by adwaita.

# Screenshots
Here's an example of a french translation which I'd love to share once i18n is fully integrated to Money:
Screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -334,6 +334,9 @@ ASALocalRun/
 # Local History for Visual Studio
 .localhistory/
 
+# Gettext model
+*.pot
+
 # Prerequisites
 *.d
 
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,7 @@ sqlitecpp = dependency('SQLiteCpp')
 boost = dependency('boost')
 
 subdir('src')
+subdir('po')
 
 executable('org.nickvision.money', sources, dependencies: [threads, adwaita, jsoncpp, dl, sqlitecpp, boost], install: true)
 install_data(resources, install_dir: 'share/icons/hicolor/scalable/apps')
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -8,4 +8,5 @@ src/ui/views/accountview.cpp
 src/ui/views/groupdialog.cpp
 src/ui/views/mainwindow.cpp
 src/ui/views/preferencesdialog.cpp
+src/ui/views/shortcutsdialog.cpp
 src/ui/views/transactiondialog.cpp
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -1,0 +1,11 @@
+src/controllers/accountviewcontroller.cpp
+src/models/transaction.cpp
+src/ui/controls/comboboxdialog.cpp
+src/ui/controls/entrydialog.cpp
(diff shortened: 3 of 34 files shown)


nlogozzo:
Bless your soul for putting this all together! I'm currently working on implementing #2 at the moment but once I have time I will look at this and merge it in.

nlogozzo:
I may not be able to merge this directly as I want to implement this in my application skeleton (https://github.com/nlogozzo/NickvisionApplication) so I can apply it to all my other applications.

nlogozzo:
Should i add "en_US" to LINGUAS since the application is coded in that language or no?

nlogozzo:
I also get a warning when I try to show a toast with a file path:
`m_sendToastCallback(StringHelpers::format(_("Folder Opened: %s"), m_folderPath));` 
gives 
`(org.nickvision.application:2): Gtk-WARNING **: 01:28:49.364: Failed to set text 'Folder Opened: \xb0\x83\xces\x94U' from markup due to error parsing markup: Error on line 1 char 30: Invalid UTF-8 encoded text in name — not valid “Folder Opened: \xb0\x83\xces\x94U”`

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2022.11.0-beta2

zothma:
> Should i add `en_US` to `po/LINGUAS` since the application is coded in that language or no?

I see what you want, but it's actually useless. Adding a language to po/LINGUAS generates a new .po file for translation. You'd then translate from the original language (American English) to the target language (American English again)...

However, it may be interesting to add the "en_GB" if you have some differences with British English.

nlogozzo:
Well translation support is here so feel free to create a PR with the French translation whenever:)

securearth:
Hindi translation would be soon available

PR #10, opened 2022-11-06
Translation/french

zothma:
# Added translation for the french language
This translation is however incomplete, as stated in #9, because the about page is generated by libadwaita


nlogozzo:
Is there anything we can do about AdwAboutDialog?

zothma:
I went and checked libadwaita source code, it turns out they have implemented translation but it doesn't happen to work on my computer. I won't open an issue here, I'll try and see what's wrong by my side

PR #12, opened 2022-11-06
Add Italian translation

bordam:
If what I requested in #1 gets implemented, I'll update the translation.


nlogozzo:
Please remove the .pot file as it is not part of the translation. 
If you would like, you can add the command to generate the .pot file in a new section in the README to make it easier for new translators. However, the actual .pot file should not be included in the project.

securearth:
@nlogozzo i would do that to readme

securearth:
Regarding the pot file (you said it is not part of translation), it is a general practice to provide the .pot file in the repository itself (i have seen in a lot of gnome apps), but what i suggest is that the app is still in development, so let it be like that and not include the .pot file, once it gets stable, and the strings in translations are stable, then we would include a .pot file

nlogozzo:
Maybe we can add it once it gets stable. However, @zothma and I have seen many gnome apps that don't include the .pot file. I think it is best to provide the command to generate it in the readme, to ensure the translator always has the most up-to-date version of the .pot file on their end.

nlogozzo:
Either way, for now it should be removed from this PR.

zothma:
> Regarding the pot file (you said it is not part of translation), it is a general practice to provide the .pot file in the repository itself (i have seen in a lot of gnome apps)
@securearth can you link us to a gnome project that does include it? As for now I haven't seen it but I haven't looked to all gnome projects, and I may be wrong thinking it is rather general practice to not include the file

securearth:
> > Regarding the pot file (you said it is not part of translation), it is a general practice to provide the .pot file in the repository itself (i have seen in a lot of gnome apps)
> 
> @securearth can you link us to a gnome project that does include it? As for now I haven't seen it but I haven't looked to all gnome projects, and I may be wrong thinking it is rather general practice to not include the file

Maybe I was wrong when I said 'most' gnome projects use it. Core gnome apps using their own platform called l10n.gnome.org (called Damned Lines), but some of gnome circle apps (basically the community apps) do provide the `.pot` file, either in the base directory or either in `po/` directory.

[https://github.com/GustavoPeredo/Font-Downloader/tree/master/po](https://github.com/GustavoPeredo/Font-Downloader/tree/master/po) (gives messages.po instead of .pot file but it is the same thing)

[https://github.com/Huluti/Curtail/tree/master/po](https://github.com/Huluti/Curtail/tree/master/po)

[https://github.com/SeaDve/Kooha](https://github.com/SeaDve/Kooha)

[https://github.com/dialect-app/po/tree/d7abf9dc4d2b6e153e6afeb2f1e6421fb94ce8f4](https://github.com/dialect-app/po/tree/d7abf9dc4d2b6e153e6afeb2f1e6421fb94ce8f4)

nlogozzo:
I think it's best to keep our app in line with core GNOME apps, which usually don't include .pot

bordam:
@nlogozzo np, I'll remove the pot file and add instructions in the readme.

nlogozzo:
@bordam you can just remove the .pot file. A PR was merged made by @securearth that included instructions to the Readme :)

nlogozzo:
Please update the translation to the new translation context seen in https://github.com/nlogozzo/NickvisionMoney/pull/17

bordam:
I'll just make a new PR at this point

PR #17, opened 2022-11-06
Support for contextual translation and small gettext corrections

zothma:
# What?
Added support for contextual translation, that means translating a same English word/sentence in several different words/sentences depending on the context.

# Why?
As explained in this [issue comment](https://github.com/nlogozzo/NickvisionMoney/issues/1#issuecomment-1304806778), one same word has to be translated differently depending on where it is in the UI.

# How?
This version adds the `pgettext` function in **src/helpers/contextualtranslation.hpp**, which allows to add a context as first argument. This context is then added in the po file with the `msgctxt` property automatically. Several entries can now have the same id but are differentiated by their context.

# Small tweaks
This version also takes care of :
 - correcting the ["About Money" issue comment](https://github.com/nlogozzo/NickvisionMoney/issues/1#issuecomment-1304829083)
 - correcting the [missing translation support](https://github.com/nlogozzo/NickvisionMoney/issues/1#issuecomment-1304813257) in file dialogs
 - updating french translation


nlogozzo:
If you don't mind, I wouldn't add these new functions to a separate `contextualtranslation.hpp` file. Could you combine it with `translation.hpp`?

zothma:
That's what I first did, but it then throws a lot of warnings because the function is not used in every file depending on "translation.hpp"

nlogozzo:
You'd have to split it into a .hpp and .cpp file for C++ to not throw multiple definition warnings.

Try implementing it like this:


`src/helpers/translation.hpp`
```cpp
#pragma once

#include <libintl.h>
#include <locale.h>

#define _(String) gettext(String)

const char* pgettext_aux(const char* msg_ctxt_id, const char* msgid);

#define GETTEXT_CONTEXT_GLUE "\004"
#define pgettext(Ctxt, String) pgettext_aux(Ctxt GETTEXT_CONTEXT_GLUE String, String)
```

`src/helpers/translation.cpp`
```cpp
#include "translation.hpp"

const char* pgettext_aux(const char* msg_ctxt_id, const char* msgid)
{
    const char* translation{ dcgettext(GETTEXT_PACKAGE, msg_ctxt_id, LC_MESSAGES) };
    if (translation == msg_ctxt_id)
    {
    	return msgid;
    }
    return translation;
}
```

(Don't forget to add the translation.cpp to the src/meson.build)

zothma:
I'm pretty sure I added something like that and I always ended up having the linker not recognizing my function... And now for some obscure reason it works... I committed these changes

nlogozzo:
It might have been because you put the `#define` above the aux method which would cause it to not see the aux method since it wasn't defined before. But who knows 🤷

nlogozzo:
Just one minor thing...could you move the `translation.cpp` below the `translation.hpp` in the `src/meson.build` file. Should be .hpp first then .cpp

zothma:
Done

nlogozzo:
Beautiful! Thank you!

PR #21, opened 2022-11-06
Fix poeditor link

bordam:
Since poedit.com does not exist, I assume you mean poeditor.com. Or maybe you mean poedit the application (poedit.net)?


nlogozzo:
I think it is supposed to be `poedit.net` if you could change it to that.

bordam:
Ok, done 👍

PR #24, opened 2022-11-07
support for translator credits, updated po files for new 'translator-credits' string

securearth:
Most of gnome apps provide translator credits in the about section of app, and the new libadwaita about dialog also has support for that. Basically a new line is added in `src/ui/views/mainwindow.cpp`, which also adds a new string in the po files called "translator-credits" and the translators put their name(s) and website/email (as described in [https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AboutWindow.html](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AboutWindow.html). )

If the app is operated in English, it does not show any translator credits, but if it is opened in some other locale, it displays translator credits in credits section of about dialog like this:

Screenshot from 2022-11-07 23-21-04

This pull request also ran the `meson compile org.nickvision.money-update-po` which updated the existing `.po` files, so you may ask the existing translators to maybe put their credits here in this discussion thread, so that you maybe update them all at one time (instead of every translator creating a new pull for their credits)

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: org.nickvision.money\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-11-06 18:53+0100\n"
+"POT-Creation-Date: 2022-11-07 23:15+0530\n"
 "PO-Revision-Date: 2022-11-06 10:02+0100\n"
 "Last-Translator: Enzo MAROS <enzo.maros@gmail.com>\n"
 "Language: fr\n"
@@ -33,31 +33,31 @@ msgstr "%d transaction importées depuis le CSV"
 msgid "None"
 msgstr "Aucun"
 
-#: src/models/transaction.cpp:55 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:55 src/ui/views/transactiondialog.cpp:56
 msgid "Never"
 msgstr "Jamais"
 
-#: src/models/transaction.cpp:59 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:59 src/ui/views/transactiondialog.cpp:56
 msgid "Daily"
 msgstr "Quotidienne"
 
-#: src/models/transaction.cpp:63 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:63 src/ui/views/transactiondialog.cpp:56
 msgid "Weekly"
 msgstr "Hebdomadaire"
 
-#: src/models/transaction.cpp:67 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:67 src/ui/views/transactiondialog.cpp:56
 msgid "Monthly"
 msgstr "Mensuelle"
 
-#: src/models/transaction.cpp:71 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:71 src/ui/views/transactiondialog.cpp:56
 msgid "Quarterly"
 msgstr "Trimestrielle"
 
-#: src/models/transaction.cpp:75 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:75 src/ui/views/transactiondialog.cpp:56
 msgid "Yearly"
 msgstr "Annuelle"
 
-#: src/models/transaction.cpp:79 src/ui/views/transactiondialog.cpp:57
+#: src/models/transaction.cpp:79 src/ui/views/transactiondialog.cpp:56
 msgid "Biyearly"
 msgstr "Semestrielle"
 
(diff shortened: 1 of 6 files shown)


nlogozzo:
@zothma What do you think?

zothma:
Often, there might be several translators, because someone added a correction to what was already made for example... However with that implementation, the last translator may only be shown (if he changes the entry in the po file)... Wouldn't it be better, if translator credits were added in money, to show them all, like this picture?

Capture d’écran du 2022-11-07 19-45-17

Here, it also features adding the email, but it may not be necessary

nlogozzo:
I wouldn't mind manually adding translators to a list like I already to for designers, artists, etc...
I think this would be the best approach.

zothma:
The problem by manually adding is that it gets all translators, and what both the screenshot and @securearth implement is only showing the translators from your language (and shows none if you're in the original version)

nlogozzo:
Ah ok I see. Could we use the singular "translator-credits" line in the .po files and have each person's info on a new line. Then in the software, split that string up by `\n` and create a list from that to show in the about dialog.

securearth:
> That woud be a good way to do it, I just saw it done in the project I took the screenshot from

Can you please provide a link of that project?

nlogozzo:
I will take care of implementing this now. It is going to require me to do some C++ magic.

zothma:
It actually seems that when putting "\n", the string is automatically split into multiple rows! My bad
That means you can have multiple creators as credit AND getting the email address thanks to the power of adwaita

nlogozzo:
> It actually seems that when putting "\n", the string is automatically split into multiple rows! 

So does that mean we don't need to do anything extra?

zothma:
No extra is needed, I have to admit I didn't know libadwaita would take care of that for us

zothma:
Apart from maybe adding the credits for existing translators instead of waiting for them to do a PR again

bordam:
This is my tag: Mattia Borda <mattiagiovanni.borda@icloud.com>
I can make a separate PR if you wish.

bordam:
@nlogozzo I don't know if this is what you asked, but if you want to display all translators of all languages in the credits section, ru translator @fsobolev might know how to do it since his app behaves this way.

nlogozzo:
> @nlogozzo I don't know if this is what you asked, but if you want to display all translators of all languages in the credits section, ru translator @fsobolev might know how to do it since his app behaves this way.

The only reason I like this approach better is to 
1. give credit to everyone's work, even if someone isn't using the specific language translation  
2. to prevent everyone from having to open new PRs just for their name tags.

fsobolev:
> @nlogozzo I don't know if this is what you asked, but if you want to display all translators of all languages in the credits section, ru translator @fsobolev might know how to do it since his app behaves this way.

In my app translators are added by hand.
```
about = Adw.AboutWindow(transient_for=self.props.active_window,
                                application_name='Time Switch',
                                ...
                                translator_credits='\n'.join([
                                'Irénée Thirion (French) <irenee.thirion@e.email>',
                                'Albano Battistella (Italian) <albano_battistella@hotmail.com>',
                                'Åke Engelbrektson (Swedish) <eson@svenskasprakfiler.se>'
                                ]))
```

I was just thinking today that it would be good to change it somehow, and my idea was exactly this:
> Could we use the singular "translator-credits" line in the .po files and have each person's info on a new line.

securearth:
> Nothing extra is needed, I have to admit I didn't know libadwaita would take care of that for us

Adwaita had this functionality build in of displaying translator of that particular locale, but I did not know that it could also split it across and make a list (because when I opened some gnome apps in my language, there were multiple translators, but they were separated by commas and not newline, so they were being dispalyed a single line)

securearth:
> > @nlogozzo I don't know if this is what you asked, but if you want to display all translators of all languages in the credits section, ru translator @fsobolev might know how to do it since his app behaves this way.
> 
> The only reason I like this approach better is to
> 
> 1. give credit to everyone's work, even if someone isn't using the specific language translation
> 2. to prevent everyone from having to open new PRs just for their name tags.

Its now completely your choice what to do

securearth:
If you want to merge this pull request, a change has to be done, when i refreshed the po files, the dutch translation did not exist, but it got merged recently, so would have to refresh for that as well

nlogozzo:
I don't think I'll be merging this. Still haven't decided on the best approach yet, but either way translators will get the credit they deserve :)

fsobolev:
I was working on improving translator credits for my app. My goal was to keep the behavior of always showing all translators, but make it relatively easy for translators to add their names. So this is how it works:
* `po` folder now contains CREDITS.json, listing all languages and translators
* the app sorts it alphabetically, by languages first, then by names, so the original order in CREDITS.json doesn't matter
* no modifications in `*.po` files needed at all

@nlogozzo if it's suitable for you, you can try to reimplement it. https://github.com/fsobolev/timeswitch/commit/fb4bd2954db3a8ec0b2086fdd47e687fedcd7ad4

PR #27, opened 2022-11-11
Removed deprecated Gtk.StyleContext

zothma:
# What
Removed deprecated functions

# Why
Since version 4.9.1 of GTK, the Gtk.StyleContext class has been deprecated in favor of the new styling functions in the Gtk.Widget class. See [GTK changelog](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/NEWS).

# How
The new method being used for adding a style is `gtk_widget_add_css_class`, and the one for removing a style is `gtk_widget_remove_css_class`. They have a much simpler syntax than the old Gtk.StyleContext. Each file using styling has been edited to use the new corresponding method.


nlogozzo:
Thanks for this!

PR #32, opened 2022-11-13
modify 'Recents' to _('Recents') and update hindi translation

securearth:
You forgot to mark the 'Recents' string for translation by enclosing in _ function. I did that, and updated all po files to include that and also updated Hindi translation.

### DIFF
--- a/po/hi.po
+++ b/po/hi.po
@@ -1,16 +1,18 @@
 # Hindi translation for Nickvision Money
 # Copyright (C) 2022 Nick Logozzo
 # This file is distributed under the same license as the org.nickvision.money package.
+# Translators:
 # Hemish <hemish04082005@gmail.com>, 2022.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: org.nickvision.money\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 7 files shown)


po/hi.po
```diff
@@ -222,52 +224,62 @@ msgstr "विवरण (खाली)"
...
+#: src/ui/views/mainwindow.cpp:41
+msgid "New"
+msgstr "नया"
 
-#: src/ui/views/mainwindow.cpp:29 src/ui/views/shortcutsdialog.cpp:98
-msgid "Close Account"
-msgstr "खाता बंद करें"
+#: src/ui/views/mainwindow.cpp:44
+msgid "New Account (Ctrl+N)"
+msgstr "नया खाता (Ctrl+G)"
```
fsobolev, line 237: @heymisphere keyboard shortcut is wrong.

heymisphere: ah sorry, would correct it now, i know what caused it. The string was previously 'New' which used to refer to group, but due to fuzzy function, 'new' is part of 'new account' so it just kept it like that, and though i changed it to account, i forgot to change the keyboard shorcut

Issue #15, opened 2022-11-06
[Bug] Importing from CSV file containing non-ASCII characters causes the application to crash


fsobolev:
As I mentioned in #14, I experience crashes when importing from CSV. I found out that it only happens if the CSV file contains non-ASCII characters (cyrillic in my case).

CSV file:
```
ID,Date,Description,Type,RepeatInterval,Amount,Group
1,2022-11-06,Тестовая транзакция,0,0,99,1
2,2022-11-06,Покупка,1,0,12,-1
```
Output:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

fsobolev:
Sorry, I forgot to mention that if after restart I open the account I was importing into, the first transaction is actually imported, but not the second one.

nlogozzo:
Mine doesn't crash, but it only imports the second transaction. 🤔🤔
![image](https://user-images.githubusercontent.com/17648453/200192380-3e01c3c4-680c-4773-9961-c58976e9b1c4.png)

fsobolev:
I see you have Test group. If I add group first, then import CSV, the result is the same as yours.
Crash happens when trying to import to new, absolutely clean account.

By the way, the crash also happens if I try to import CSV that has more transactions than the account. Even for files containing only ASCII-characters (both CSV and .nmoney).
If I try to import CSV that has less transactions than the account, nothing happens (notification shows that 0 transactions imported).

nlogozzo:
Oh yes I need to fix the import and export csv to support importing and exporting groups. Right now only transactions are handled, and if the group doesn't exist, it won't import the transaction.

nlogozzo:
@zothma is working on a fix for the CSV import/export system :) It should be finished soon!

nlogozzo:
Fixed by #39

PR #39, opened 2022-11-14
Better handling of groups in .nmoney and .csv files

zothma:
# What
This pull request focuses on a better integration of groups in the nmoney database files and csv files.

# Why
What motivated the change in the nmoney file is that the balance was hardcoded, while it simply is a sum of all the transactions' amounts, which SQL could easily handle. There was clearly a kind of indirect data duplication, which is now removed. As for the CSV file, it did not allow to export groups so far.

# How
The new `groups` table in nmoney files no longer has a balance column. Instead, a SQL request is made to do the calculation of the balance, as seen in the new `Account::updateGroupAmounts` method. The column will be automatically deleted on old files to match with the new file version.

In the CSV file, each transaction comes with its gid, group name and group description, so that the group can be added if it does not exist. This means that a group will surely appear more than once, but this limitation is due to CSV files being having only a single table, not two tables as in SQL.

@fsobolev

### DIFF
--- a/src/models/account.cpp
+++ b/src/models/account.cpp
@@ -383,7 +383,7 @@ int Account::importFromCSV(const std::string& path)
         {
             //Separate fields by ,
             std::vector<std::string> fields{ split(line, ",") };
-            if(fields.size() != 7)
+            if(fields.size() != 9)
             {
                 continue;
             }
@@ -455,21 +455,19 @@ int Account::importFromCSV(const std::string& path)
             }
             //Get Group Id
             int gid{ 0 };
-            if(gid != -1)
+            try
             {
-                try
-                {
-                    gid = stoui(fields[6]);
-                }
-                catch(...)
-                {
-                    continue;
-                }
-                if(getGroupById(gid) != std::nullopt)
-                {
-                    continue;
-                }
+                gid = stoui(fields[6]);
             }
+            catch(...)
+            {
+                continue;
+            }
+            //Get Group Name
+            const std::string& groupName{ fields[7] };
+            //Get Group Description
+            const std::string& groupDescription{ fields[8] };
+
             //Add Transaction
             Transaction transaction{ id };
             transaction.setDate(date);
@@ -479,9 +477,19 @@ int Account::importFromCSV(const std::string& path)
             transaction.setAmount(amount);
             transaction.setGroupId(gid);
             addTransaction(transaction);
+
+            //Add Group if needed
+            if (getGroupById(gid) == std::nullopt && gid != -1) {
+                Group group{ (unsigned int) gid };
+                group.setName(groupName);
+                group.setDescription(groupDescription);
+                addGroup(group);
+            }
+
             imported++;
         }
     }
+    updateGroupAmounts();
     return imported;
 }
 
--- a/src/models/account.cpp
+++ b/src/models/account.cpp
@@ -482,17 +482,11 @@ int Account::importFromCSV(const std::string& path)
 
 void Account::updateGroupAmounts()
 {
-    for(std::pair<const unsigned int, Group>& pair : m_groups)
+    //Query for balance in SQL
+    SQLite::Statement qryGetGroupsBalance{ *m_db, "SELECT g.id, CAST(COALESCE(SUM(IIF(t.type=1, -t.amount, t.amount)), 0) AS TEXT) FROM transactions t RIGHT JOIN groups g on g.id = t.gid GROUP BY g.id;" };
+    while(qryGetGroupsBalance.executeStep())
     {
-        pair.second.setBalance(0);
-    }
-    for(const std::pair<const unsigned int, Transaction>& pair : m_transactions)
-    {
-        if(pair.second.getGroupId() != -1)
-        {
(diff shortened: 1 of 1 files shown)


src/models/account.cpp
```diff
@@ -36,8 +36,13 @@ unsigned int stoui(const std::string& str, size_t* idx = nullptr, int base = 10)
 Account::Account(const std::string& path) : m_path{ path }, m_db{ std::make_shared<SQLite::Database>(m_path, SQLite::OPEN_READWRITE | SQLite::OPEN_CREATE) }
 {
     //Load Groups
-    m_db->exec("CREATE TABLE IF NOT EXISTS groups (id INTEGER PRIMARY KEY, name TEXT, description TEXT, balance TEXT)");
-    SQLite::Statement qryGetAllGroups{ *m_db, "SELECT * FROM groups" };
+    m_db->exec("CREATE TABLE IF NOT EXISTS groups (id INTEGER PRIMARY KEY, name TEXT, description TEXT)");
+    try
+    {
+        m_db->exec("ALTER TABLE groups DROP COLUMN balance");
```
nlogozzo, line 42: I don't think this 
```cpp
try
{
     m_db->exec("ALTER TABLE groups DROP COLUMN balance");
}
catch(...) {  }
```
is needed, since there is no `Groups` table in the V2022.10.0 stable and I'm ok with breaking API between betas.

src/models/account.cpp
```diff
@@ -474,25 +477,29 @@ int Account::importFromCSV(const std::string& path)
             transaction.setAmount(amount);
             transaction.setGroupId(gid);
             addTransaction(transaction);
+
+            //Add Group if needed
+            if (getGroupById(gid) == std::nullopt && gid != -1) {
```
nlogozzo, line 482: Could you make if statements like this:
```cpp
if
{
}
```
instead of 
```cpp
if {
}
```
Just to keep styles consistent :)

src/models/account.cpp
```diff
@@ -360,7 +359,13 @@ bool Account::exportAsCSV(const std::string& path)
             file << static_cast<int>(pair.second.getType()) << ",";
             file << static_cast<int>(pair.second.getRepeatInterval()) << ",";
             file << pair.second.getAmount() << ",";
-            file << pair.second.getGroupId() << "\n";
+            file << pair.second.getGroupId() << ",";
+            if (pair.second.getGroupId() != -1) {
```
nlogozzo, line 363: See style comment below.

src/models/account.cpp
```diff
@@ -474,25 +477,29 @@ int Account::importFromCSV(const std::string& path)
...
+    while(qryGetGroupsBalance.executeStep())
     {
-        if(pair.second.getGroupId() != -1)
-        {
-            Group group{ m_groups.at(pair.second.getGroupId()) };
-            group.setBalance(group.getBalance() + (pair.second.getType() == TransactionType::Income ? pair.second.getAmount() : (pair.second.getAmount() * -1)));
-            updateGroup(group);
-        }
+        int index{ qryGetGroupsBalance.getColumn(0).getInt() };
+        m_groups[index].setBalance(boost::multiprecision::cpp_dec_float_50(qryGetGroupsBalance.getColumn(1).getString()));
```
nlogozzo, line 503: When dealing with `std::map` types please use `.at()` instead of `[]`
So 
`m_groups[index]` --> `m_groups.at(qryGetGroupsBalance.getColumn(0).getInt()).setBalance()...` 
(you can save the allocation of creating an `index` variable as well.

nlogozzo (changes requested):
Looks pretty good to me so far! The requested changes are just some little style things.
I would just like to do some runtime testing of the PR before I merge, but overall great work :)

src/models/account.cpp
zothma: For the allowation of the index, I feel like it is more readable, isn't it? The line would go be quite long if it was added in the `at()`

nlogozzo: It is more readable but we save an allocation of 4 byte :)) 
I've learned to write for performant code instead of readable working with C++ haha.

nlogozzo: You could also change the query variable name to just qryGroup to save some space.

nlogozzo:
This should also fix #15

src/models/account.cpp
zothma: I was just saying, easier to maintain haha, especially for only 4 bytes

nlogozzo:
Thanks for the changes :) like I said, I just want to do some testing when I get a chance and then will merge.

fsobolev:
Thanks @zothma! I don't think I can help with the code if some places to improve it will be found, but I will definitely test it, especially I want to check how it affects #15

fsobolev:
I confirm that importing/exporting CSV file with cyrillic symbols now works, so #15 will be fixed. That's awesome! 🎉 

There are still some CSV import issues, but I doubt they can be fixed in this PR. I will open a new issue.

nlogozzo (approved):

nlogozzo:
CSV definitely works a lot better! Thanks for this!

PR #57, opened 2022-11-17
Updated french translation

zothma:
Updated french translation


zothma:
This new commit should overwrite the existing po file

Issue #48, opened 2022-11-16
Plural forms


milotype:
@nlogozzo If possible please use plural forms in strings where counts appear. Several languages need more than one plural form. IMO the plural form should be used, even if a singular form of the string will never get used.

Example **without** plular forms wich is currently used:
```
#: src/controllers/accountviewcontroller.cpp:78
#, c-format
msgid "Imported %d transactions from CSV."
msgstr "Uvezene transakcije iz CSV datoteke: %d."

```
Example **with** plular forms (for Croatian):
```
#: src/controllers/accountviewcontroller.cpp:78
#, c-format
msgid "Imported %d transaction from CSV."
msgid_plural "Imported %d transactions from CSV."
msgstr[0] "Uvezena je %d transakcija iz CSV datoteke"
msgstr[1] "Uvezene su %d transakcije iz CSV datoteke"
msgstr[2] "Uvezeno je %d transakcija iz CSV datoteke"
```

I'm not a programmer, so I don't know if `c-format` or `python-format` have to be used …

nlogozzo:
@zothma

Issue #60, opened 2022-11-17
[UI/UX] Account View Redesign


fsobolev:
This is a mockup for new Account View. It's not fully finished: I didn't find a way to make grid adaptive and didn't implement all the logic to make new features work. Yet. So I can't show how it will look with different window sizes, but at least you can now see how it looks by default and so we can start conversation to improve it :blush:

New features:
* Adaptive design: left pane will hide when the window is too narrow (like in Nautilus or Amberol), and the grid of transactions will change its number of columns depending on the width;
* Filter by income/expense;
* Filter by groups;
* Filter by dates;
* Transactions have icons: a user will select an icon from pre-defined list and its color (similar functionality can be seen in [Firefox Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/) and [Paper](https://posidon.io/paper/))

Also remove button will be moved from transactions rows to transaction edit dialog.

Screenshots:

![](https://i.imgur.com/8X8hwbk.png)
![](https://i.imgur.com/ZuwmEfl.png)

Things to think about:
1. Order of button in headerbar: should it be Account Menu Button first and then Toggle Pane Button, or vice versa?
2. Should we allow user to select any color for icons or use pre-defined list that will change depending on the app theme (light/dark)? On screenshots you can see that icons have different colors depending on theme, but that's because I used pre-defined Adwaita styles using `add_css_class(...)`. Some arbitrary colors will look bad with dark or light theme.
3. Do we need to do something with calendar? It's just a `Gtk.Calendar` widget and it looks... not bad, but not "adwait-y" enough.

P.S. Just like with new Start Screen, I made this mockup using Python, but of course I will work on actually implementing it in the app as soon as we discuss the details.

nlogozzo:
1. I think the order of the buttons in the header bar are fine.
2. Maybe instead of a whole set of icons...what if we did colored dots. Similar to how transaction 1 has a yellow dot, what if we just allowed the user to chose the color of the dot (kind of like how tags in a file manager work)
3. Try adding the `card` css class to the calendar widget (that's usually what I do when a widget doesn't look adwaita-y enough)

nlogozzo:
But I love this design! I want to stay up all night to implement this now haha

nlogozzo:
@zothma

fsobolev:
[...]

PR #62, opened 2022-11-18
Support for translating plurals

zothma:
# What
Support for translating plurals, and updated french translation to that particular behavior.

# Why
As explained in #48, when working with numbers, languages don't behave the same way. While English only has two forms, singular and plural, other languages like Croatian may have up to two plurals depending on the amount before. It is thus important for each language to be able to translate to the right word thanks by taking into account the number it is preceded by.

# How
The only string that must be changed is when importing CSV files, in **src/controllers/accountviewcontroller.cpp**

I used the `ngettext` function which takes as arguments the English singular form, the plural form, and the number actually written by `%d`,. Gettext will then regenerate each PO file, find how many different plural forms have to be created thanks to the `Plural-Form` header, and write as much entry as needed. To find the correct plural form to use, see [Gettext official documentation](https://www.gnu.org/software/gettext/manual/gettext.html#index-specifying-plural-form-in-a-PO-file).

*For this change to apply to all languages, PO files have to be regenerated by meson. It is preferable that each translator adds his plural form header first and regenerates his own file rather than regenerating them all at once right now.*

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -7,13 +7,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: org.nickvision.money\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-11-17 10:33+0100\n"
-"PO-Revision-Date: 2022-11-06 10:02+0100\n"
+"POT-Creation-Date: 2022-11-18 22:06+0100\n"
+"PO-Revision-Date: 2022-11-18 22:09+0100\n"
 "Last-Translator: Enzo MAROS <enzo.maros@gmail.com>\n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=n>1;\n"
 
 #: src/controllers/accountviewcontroller.cpp:63
 msgid "Exported account to CSV successfully."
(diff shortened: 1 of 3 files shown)


zothma:
@milotype can you confirm that this PR does work for you?

milotype:
@zothma Can you please prepare the .po file for Croatian. I'm not a programmer and don't know how to regenerate the .po file by meson.

I'd also like to suggest that you provide a current .pot file in the po-folder. That helps when translating into new laguages and can be used to update existing translated .po files.

zothma:
> @zothma Can you please prepare the .po file for Croatian. I'm not a programmer and don't know how to regenerate the .po file by meson.
> 
> I'd also like to suggest that you provide a current .pot file in the po-folder. That helps when translating into new laguages and can be used to update existing translated .po files.

Did it! If you want for next time, you can run `meson compile org.nickvision.money-update-po` inside your build directory to achieve that. Don't hesitate and read the README too, someone took time to explain it all in there 😉

milotype:
@zothma Thanks for the new hr-file! 
Should I simply sent a new PR from my fork?

As mentioned I'm not a programmer and don't know anything about running meson ... :-(

milotype:
@zothma OK, so I've just sent a new PR #63 for the Croatian PO file with plural forms.

nlogozzo:
Thanks for this <3!

nlogozzo:
I wouldn't ping all translators to update their transactions just yet. After we finish #60 (which is almost there), translators will need to update everything.

PR #100, opened 2022-11-28
Default Colors Settings

fsobolev:
Changing colors settings doesn't affect existing transactions, even if they use default color. For transfer transaction, `purple_2` is used because:
1. It looks good with both light and dark theme;
2. It's close enough to the default blue, but different enough to be easy to notice (at least for non-color-blind users);
3. It doesn't have a fixed meaning in Adwaita design.

New strings use gettext, but I didn't regenerate PO files since we'll do it later anyway when more strings will be added.

### DIFF
--- a/src/controllers/transactiondialogcontroller.cpp
+++ b/src/controllers/transactiondialogcontroller.cpp
@@ -7,7 +7,7 @@ using namespace NickvisionMoney::Controllers;
 using namespace NickvisionMoney::Helpers;
 using namespace NickvisionMoney::Models;
 
-TransactionDialogController::TransactionDialogController(unsigned int newId, const std::map<unsigned int, Group>& groups, const std::locale& locale, const std::string& defaultColor) : m_response{ "cancel" }, m_locale{ locale }, m_transaction{ newId }, m_groups{ groups }, m_defaultColor{ defaultColor }
+TransactionDialogController::TransactionDialogController(unsigned int newId, const std::map<unsigned int, Group>& groups, Configuration& configuration) : m_response{ "cancel" }, m_transaction{ newId }, m_groups{ groups }, m_configuration{ configuration }
 {
     for(const std::pair<const unsigned int, Group>& pair : m_groups)
     {
@@ -17,7 +17,7 @@ TransactionDialogController::TransactionDialogController(unsigned int newId, con
     m_groupNames.insert(m_groupNames.begin(), _("None"));
 }
 
-TransactionDialogController::TransactionDialogController(const Transaction& transaction, const std::map<unsigned int, Group>& groups, const std::locale& locale, const std::string& defaultColor) : m_response{ "cancel" }, m_locale{ locale }, m_transaction{ transaction }, m_groups{ groups }, m_defaultColor{ defaultColor }
+TransactionDialogController::TransactionDialogController(const Transaction& transaction, const std::map<unsigned int, Group>& groups, Configuration& configuration) : m_response{ "cancel" }, m_transaction{ transaction }, m_groups{ groups }, m_configuration{ configuration }
 {
     for(const std::pair<const unsigned int, Group>& pair : m_groups)
     {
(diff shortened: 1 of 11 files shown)


src/controllers/transactiondialogcontroller.hpp
```diff
@@ -148,9 +155,9 @@ namespace NickvisionMoney::Controllers
 
 	private:
 		std::string m_response;
-		const std::locale& m_locale;
 		NickvisionMoney::Models::Transaction m_transaction;
 		std::map<unsigned int, NickvisionMoney::Models::Group> m_groups;
 		std::vector<std::string> m_groupNames;
+		NickvisionMoney::Models::Configuration& m_configuration;
```
nlogozzo, line 161: Can you move this variable above `m_transaction` where `m_locale` used to be.

src/controllers/transactiondialogcontroller.cpp
```diff
@@ -7,7 +7,7 @@ using namespace NickvisionMoney::Controllers;
 using namespace NickvisionMoney::Helpers;
 using namespace NickvisionMoney::Models;
 
-TransactionDialogController::TransactionDialogController(unsigned int newId, const std::map<unsigned int, Group>& groups, const std::locale& locale) : m_response{ "cancel" }, m_locale{ locale }, m_transaction{ newId }, m_groups{ groups }
+TransactionDialogController::TransactionDialogController(unsigned int newId, const std::map<unsigned int, Group>& groups, Configuration& configuration) : m_response{ "cancel" }, m_transaction{ newId }, m_groups{ groups }, m_configuration{ configuration }
```
nlogozzo, line 10: Move `m_configuration{ configuration }` to before `m_transaction{ newId }`

src/controllers/transactiondialogcontroller.cpp
```diff
@@ -17,7 +17,7 @@ TransactionDialogController::TransactionDialogController(unsigned int newId, con
     m_groupNames.insert(m_groupNames.begin(), _("None"));
 }
 
-TransactionDialogController::TransactionDialogController(const Transaction& transaction, const std::map<unsigned int, Group>& groups, const std::locale& locale) : m_response{ "cancel" }, m_locale{ locale }, m_transaction{ transaction }, m_groups{ groups }
+TransactionDialogController::TransactionDialogController(const Transaction& transaction, const std::map<unsigned int, Group>& groups, Configuration& configuration) : m_response{ "cancel" }, m_transaction{ transaction }, m_groups{ groups }, m_configuration{ configuration }
```
nlogozzo, line 20: See above comment

src/models/transaction.cpp
```diff
@@ -3,7 +3,7 @@
 
 using namespace NickvisionMoney::Models;
 
-Transaction::Transaction(unsigned int id) : m_id{ id }, m_date{ boost::gregorian::day_clock::local_day() }, m_description{ "" }, m_type{ TransactionType::Income }, m_repeatInterval{ RepeatInterval::Never }, m_amount{ 0.00 }, m_groupId{ -1 }, m_rgba{ "#3584e4" }
+Transaction::Transaction(unsigned int id) : m_id{ id }, m_date{ boost::gregorian::day_clock::local_day() }, m_description{ "" }, m_type{ TransactionType::Income }, m_repeatInterval{ RepeatInterval::Never }, m_amount{ 0.00 }, m_groupId{ -1 }, m_rgba{ "" }
```
nlogozzo, line 6: I would still provide `m_rgba{ "rgb(53,132,228)" }` in case someone forgets to call `setRGBA`

src/ui/views/preferencesdialog.cpp
```diff
@@ -25,6 +25,30 @@ PreferencesDialog::PreferencesDialog(GtkWindow* parent, const PreferencesDialogC
     adw_combo_row_set_model(ADW_COMBO_ROW(m_rowTheme), G_LIST_MODEL(gtk_string_list_new(new const char*[4]{ _("System"), _("Light"), _("Dark"), nullptr })));
     adw_preferences_group_add(ADW_PREFERENCES_GROUP(m_grpUserInterface), m_rowTheme);
     g_signal_connect(m_rowTheme, "notify::selected-item", G_CALLBACK((void (*)(GObject*, GParamSpec*, gpointer))[](GObject*, GParamSpec*, gpointer data) { reinterpret_cast<PreferencesDialog*>(data)->onThemeChanged(); }), this);
+    //Transaction Default Color Row
+    GdkRGBA transactionColor;
+    gdk_rgba_parse(&transactionColor, m_controller.getTransactionDefaultColor().c_str());
+    m_btnTransactionColor = gtk_color_button_new_with_rgba(&transactionColor);
```
nlogozzo, line 31: Loading the color should be moved to the `//Load Configuration` section of the constructor.

The buttons can be initialized with `gtk_color_button_new()` and the color can be then loaded with
```cpp
GdkRGBA transactionColor;
gdk_rgba_parse(&transactionColor, m_controller.getTransactionDefaultColor().c_str());
gtk_color_chooser_set_rgba(GTK_COLOR_CHOOSER(m_btnTransactionColor), &transactionColor);
```
in the appropriate section.

nlogozzo (changes requested):

src/models/transaction.cpp
fsobolev: Then the default color will not be set in `transactiondialog.cpp`:
```
if(!gdk_rgba_parse(&color, m_controller.getRGBA().c_str()))
    {
        gdk_rgba_parse(&color, m_controller.getTransactionDefaultColor().c_str());
    }
```

nlogozzo: Oh yes you are right sorry.

nlogozzo (approved):

PR #125, opened 2022-12-07
Fix decimal zeros (again)

### LINKED ISSUE #97: Localisation issue with real numbers
I just came across a weird behavior appearing almost every ten cents. Here is an image

Capture d’écran du 2022-11-27 19-30-41

From what I tried, the last 0 only disappear with numbers greater than $1 ($0.10, $0.20... appear correctly), every 10 cents except for 0 cents and 50 cents. So, from $0.90 to $1.50, numbers would appear like this (in euros):

Capture d’écran du 2022-11-27 20-59-46

fsobolev:
Close #97

### DIFF
--- a/src/helpers/moneyhelpers.cpp
+++ b/src/helpers/moneyhelpers.cpp
@@ -1,5 +1,4 @@
 #include "moneyhelpers.hpp"
-#include <cmath>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


nlogozzo:
Thanks for this!

PR #148, opened 2022-12-29
GNOME - Fix GResource location

fsobolev:
If the app is installed in `/opt` the gresource will not be found, so let the app first look in `/usr/share` and only then in `../../share`


nlogozzo:
You already know what I'm going to ask about Application template :))

nlogozzo:
Let me know when this is ready

fsobolev:
> You already know what I'm going to ask about Application template :))

Yeah, haha, I just want to be sure that it will work now no matter where the app is installed

nlogozzo (approved):

PR #149, opened 2022-12-30
GNOME - Add button to show/hide groups

### LINKED ISSUE #140: Hide groups
I think will be better if the Groups could be hide  or minimized on side panel.

fsobolev:
Making groups list scrollable creates some UI problems that are not easy to fix, so instead I added show/hide button as was suggested by @brunnogama. Closes #140

### DIFF
--- a/NickvisionMoney.GNOME/Views/AccountView.cs
+++ b/NickvisionMoney.GNOME/Views/AccountView.cs
@@ -236,6 +236,7 @@ public partial class AccountView
         //Button Toggle Groups
         _btnToggleGroups = Gtk.ToggleButton.New();
         _btnToggleGroups.AddCssClass("flat");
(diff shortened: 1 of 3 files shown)


nlogozzo (approved):
LGTM!
What do you think about adding an option to the configuration, similar to SortFirstToLast, that remembers whether the user has the groups hidden or not?

fsobolev:
> LGTM! What do you think about adding an option to the configuration, similar to SortFirstToLast, that remembers whether the user has the groups hidden or not?

If a user has only one account, ability to remember groups visibility is useful. But if a user has multiple accounts and want to hide groups only in some of them, remembering groups visibility would be irritating. So we either keep it as it is, or add an option in settings to let users enable remembering groups visibility (and it will be disabled by default).

Issue #150, opened 2022-12-30
Typo at /Shared/Resources/string.resx


AsBenDoge:
Line 185,

```
<data name="Greeting.Night" xml:space="preserve">
	<value>Good Morning!</value>
</data>
```

I'm assuming it's a typo as this is a duplicated data. If it's not the Russian translation has to be updated. (Italian follows English ver.)

fsobolev:
It's not a mistake and Russian translation is correct. In English you can't say "Good night" as a greeting, people say "Good morning" even if the time is 0am-6am. But in Russian it's different (and probably in some other languages too.)

PR #151, opened 2022-12-30
Add comments to greeting strings

fsobolev:
To make them obvious and avoid duplicates of #150


nlogozzo:
Could you do the same for Application :) ?

Issue #139, opened 2022-12-24
New name?


bertob:
As discussed in the Circle review, this app could benefit from a having a more unique name. Some ideas to kick things off:

- Tender
- Stacks
- Ebenezer
- Scrooge
- Dough
- Bucks
- Balance
- Overdraft
- Coins

nlogozzo:
Maybe other language names? I'm italian so:

- Portafoglio = Wallet
- Soldi = Money
- Denaro = Money
- Contanti = Cash

fsobolev:
* Coin Purse
* Porte Monnaie
* Cash Box

bertob:
Non-english words can work, but only in rare cases where the term is somewhat international. At the very least it needs to be easy to pronounce, so definitely not portafoglio :D

heymisphere:
My suggestion: "FinoMan" from "Finance Manager"

nlogozzo:
@bertob I like `Denaro`

Comment from @fsobolev:
> "Denaro" is easy to pronounce and I guess it refers to "dinar" which is a common term in some countries, so if this will work I agree that it's great

bertob:
Denaro could work, yeah.

nlogozzo:
[Goodbye Money, Hello Denaro!](https://github.com/nlogozzo/NickvisionMoney/commit/511de12835a8022decd3f7a8f5c283a7ca3d4b68)

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.1.0-rc1

PR #152, opened 2022-12-30
Added Traditional Chinese Translation

AsBenDoge:
Sorry for #150 :(

### DIFF
--- a/NickvisionMoney.Shared/Resources/Strings.zh.resx
+++ b/NickvisionMoney.Shared/Resources/Strings.zh.resx
@@ -117,7 +117,7 @@
 
 	<!--MainWindow-->
 	<data name="NickvisionMoneyAccount" xml:space="preserve">
-		<value>Nickvision的帳戶</value>
+		<value>Nickvision的軟體--Money的帳戶</value>
 	</data>
 
 	<data name="NewAccount" xml:space="preserve">
@@ -638,6 +638,6 @@
 	</data>
 
 	<data name="AccountFileFilter.GTK" xml:space="preserve">
-		<value>帳戶 (*.nmoney)</value>
+		<value>軟體Money的帳戶 (*.nmoney)</value>
 	</data>
 </root>
--- a/NickvisionMoney.Shared/Resources/Strings.zh.resx
+++ b/NickvisionMoney.Shared/Resources/Strings.zh.resx
@@ -1,0 +1,643 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+	<resheader name="resmimetype">
+		<value>text/microsoft-resx</value>
+	</resheader>
+
+	<resheader name="version">
+		<value>2.0</value>
+	</resheader>
+
+	<resheader name="reader">
+		<value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+	</resheader>
+
+	<resheader name="writer">
+		<value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+	</resheader>
+
+	<!--Use: &#xA; for newline-->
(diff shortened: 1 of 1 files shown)


fsobolev:
> Sorry for #150 :(

Don't worry, I guess somebody would eventually asked about this anyway ☺️ And now there will be comments, so this is an improvement made thanks to you :)

NickvisionMoney.Shared/Resources/Strings.zh.resx
```diff
@@ -0,0 +1,643 @@
...
+		<value>偏好設定</value>
+	</data>
+
+	<data name="KeyboardShortcuts" xml:space="preserve">
+		<value>快捷鍵</value>
+	</data>
+
+	<!--MainWindow-->
+	<data name="NickvisionMoneyAccount" xml:space="preserve">
+		<value>Nickvision的帳戶</value>
```
fsobolev, line 120: Here "Money" is part of the application name "Nickvision Money", so IMO "Money" should be kept untranslated

NickvisionMoney.Shared/Resources/Strings.zh.resx
```diff
@@ -0,0 +1,643 @@
...
+	<data name="CloseAccount.GTK" xml:space="preserve">
+		<value>關閉帳戶</value>
+	</data>
+
+	<data name="AccountActions.GTK" xml:space="preserve">
+		<value>動作</value>
+	</data>
+
+	<data name="AccountFileFilter.GTK" xml:space="preserve">
+		<value>帳戶 (*.nmoney)</value>
```
fsobolev, line 641: And here it's probably the same situation

fsobolev (changes requested):

NickvisionMoney.Shared/Resources/Strings.zh.resx
AsBenDoge: "帳戶" means "bank account". If it translates to "Nickvision's money bank account" it'd be weird. Would "Nickvision's application 'Money' Account" be better?

NickvisionMoney.Shared/Resources/Strings.zh.resx
AsBenDoge: It's pretty confusing to say "money bank account". If we had a new name #139  these could be easily changed.

NickvisionMoney.Shared/Resources/Strings.zh.resx
fsobolev: > If it translates to "Nickvision's money bank account" it'd be weird.

I don't see how it's weird. Because it's not just "money" here, it's "Money" - the name of the app.
We're going to rename the app to Denaro soon. So it will be "Denaro Account" there. Would it be weird to translate as "Denaro bank account"? By the same logic, yes, because "denaro" means "money" in Italian. But it doesn't matter - because it's a name, and its meaning in original language isn't valid in the context. Same goes to "Money" currently.
However, I'm not a professional translator, that's my opinion as hobbyist (before software I had been translating comic books 😊 )

fsobolev: @nlogozzo I request your opinion, please

AsBenDoge: > I don't see how it's weird. Because it's not just "money" here, it's "Money" - the name of the app.

I get you but using mixed Chinese and English is often complained by users. Denaro is more clearly a name, while Money is a term. What I suggested there - specifying it as a name instead of a term, prevents that situation.

AsBenDoge: Has been updated to include the name "Money" but with "software" prefix to prevent confusion.

fsobolev (approved):

NickvisionMoney.Shared/Resources/Strings.zh.resx
nlogozzo: For now leave it as is, the name change will be implemented before the stable release so we'll finalize the translation then.

PR #155, opened 2022-12-31
Updates and pdf generation script

AsBenDoge:
**The pdf_gen.c is largely based on the example scripts at https://libharu.sourceforge.net/examples.html. It's not licensed under MIT and proper credits have to be given in README. Please consider before approving.**

The pdf_gen.c script is not finished. I can handle it if you don't mind my poorly constructed code :( I'm learning

Happy New Year!


nlogozzo:
Money is written in C# and not C, so this script serves no purpose for PDF generation.

The PDF generation is being handled by QuestPDF and occurs in the [Account class](https://github.com/nlogozzo/NickvisionMoney/blob/main/NickvisionMoney.Shared/Models/Account.cs#L583).

PR #157, opened 2022-12-31
Make manifests build from current directory

fsobolev:
GNOME Builder ignores sources lines for main module and builds the app from current directory anyway, so I didn't notice that in the default manifest sources point to specific commit, thus not allowing to test current changes.
Going to add the same fix to Application.

### DIFF
--- a/NickvisionMoney.GNOME/org.nickvision.money-gnomebuilder.json
+++ b/NickvisionMoney.GNOME/org.nickvision.money-gnomebuilder.json
@@ -40,9 +40,8 @@
             ],
             "sources": [
                 {
-                    "type": "git",
-                    "url": "https://github.com/nlogozzo/NickvisionMoney.git",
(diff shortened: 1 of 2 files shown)


nlogozzo:
I would like to keep `NickvisionMoney.GNOME/org.nickvision.money.json` as close to what is published on flathub as possible. 
So I think that should stay the way it is. But I'm all for changing the gnomebuilder one since that is used to run the app from the IDE.

fsobolev:
The problem is that to build the app from the current code using flatpak-builder this change is required. What's the point of having the manifest in the same state as in flathub repo? If a user want to build the stable/beta version of the app, the app's flathub repo can be cloned and then manifests from there can be used. It's pointless to clone the whole app source code repo to then build the app from specific release commit.

nlogozzo:
Yeah no that's true, i'll go ahead and merge!

PR #166, opened 2023-01-05
NickvisionMoney.GNOME/install_extras.sh: misc fixes

xgqt:
- break up long lines
- change the interpreter to sh (and conform to POSIX)
- make the script executable
- quote echo strings
- quote sensitive variables
- turn repetitive actions into for-loops
- use basename + pwd instead of bash-specific syntax

Signed-off-by: Maciej Barć <xgqt@gentoo.org>

### DIFF
--- a/NickvisionMoney.GNOME/install_extras.sh
+++ b/NickvisionMoney.GNOME/install_extras.sh
@@ -1,46 +1,57 @@
-#!/bin/bash
+#!/bin/sh
 
 INSTALL_PREFIX="/usr"
-if [ ! -z $1 ]
+if [ -n "${1}" ]
 then
-	INSTALL_PREFIX=$1
+    INSTALL_PREFIX="${1}"
 fi
-echo Install prefix: $INSTALL_PREFIX
+echo Install prefix: "${INSTALL_PREFIX}"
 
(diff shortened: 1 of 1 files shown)


fsobolev (approved):

fsobolev:
@nlogozzo I will later update the script in other branches and the Application

xgqt:
> Unfortunately Github doesn't remember this. I cloned you repo and the script wasn't executable, as I expected. I don't think there's a way to change this.

That is very weird... I might have messed something up on my end, git should remember executable permissions.

xgqt:
It seems to work fine actually, see:
image

fsobolev:
Hm, I thought it's something with Github, but then I guess it's something on my end. Anyway, even if I fix it for myself, if somebody with the same problem will open a PR then the script's permissions will be overwritten, so I think it's better to keep it as it is, and call it using "sh install.sh" instead of "./install.sh", just like it's already done in flatpak manifests.

xgqt:
> so I think it's better to keep it as it is, and call it using "sh install.sh" instead of "./install.sh", just like it's already done in flatpak manifests.

Yup, I also think it still should be called using `sh install_extras.sh` by other scripts.

PR #178, opened 2023-01-06
Update Italian translation

bordam:


nlogozzo (approved):

fsobolev:
@nlogozzo I noticed the release notes in metainfo are translated in this PR. But it's kinda pointless, since you clear them every release. Don't you want to start a tradition of keeping release notes history starting the first release under Denaro name? :)

fsobolev:
> > I noticed the release notes in metainfo are translated in this PR. But it's kinda pointless, since you clear them every release.
> 
> It is kind of pointless since they get cleared every release, but there's nothing wrong with it so that's why I accepted it.
> 
> > Don't you want to start a tradition of keeping release notes history starting the first release under Denaro name? :)
> 
> I've never really kept a history of release notes in the metadata file to 1. prevent it from being super long and 2. anyone can go back to the releases on the repo page and see past changes.

Okay 👍🏻

PR #198, opened 2023-01-09
Implement Sort Transaction By

### LINKED ISSUE #197: Sort by date rather than order entered
### Discussed in https://github.com/nlogozzo/NickvisionMoney/discussions/196

<div type='discussions-op-text'>

<sup>Originally posted by **jameswestman** January  9, 2023</sup>
There's an option to sort ascending or descending, but it doesn't look like there's a way to sort by date rather than the order the transactions are entered. I'd like to add a bunch of older transactions from before I started using the app, but I'd prefer they don't show up before all my other transactions that really are recent.

Thanks for this wonderful app btw!</div>

nlogozzo:
Closes #197


fsobolev:
I changed a few things for GNOME UI:
1. `_cmbSortTransactionBy` changed to `_ddSortTransactionBy` for consistency, all Gtk.DropDowns var names start with `_dd` (I think we need to create coding style guidelines that will also include variable naming, and I don't mind to use `cmb` prefix for DropDown, but for now I used what we used before);
2. DropDown is created using `NewFromStrings`, to make it shorter;
3. There are 2 boxes now: one with toggle buttons, that is inside other box with dropdown. This is required because we don't need dropdown to be affected by `linked` CSS class.

I tested that it works great, so it's ready to be merged.

fsobolev (approved):

PR #205, opened 2023-01-11
Russian translation for docs

fsobolev:
Also fixed a few mistakes in English docs, and made WinUI open correct url depending on language.


NickvisionMoney.WinUI/Views/PreferencesPage.xaml.cs
```diff
@@ -182,7 +183,14 @@ private string RemoveUrlsFromCredits(string s)
...
     /// <param name="sender">object</param>
     /// <param name="e">RoutedEventArgs</param>
-    private async void Help(object sender, RoutedEventArgs e) => await Launcher.LaunchUriAsync(new Uri($"https://htmlpreview.github.io/?https://github.com/nlogozzo/NickvisionMoney/blob/{_controller.AppInfo.Version}/NickvisionMoney.Shared/Docs/html/C/index.html"));
+    private async void Help(object sender, RoutedEventArgs e)
+    {
+        var lang = CultureInfo.CurrentCulture.ToString() switch {
+            "ru-RU" => "ru",
+            _ => "C"
+        };
+        await Launcher.LaunchUriAsync(new Uri($"https://htmlpreview.github.io/?https://raw.githubusercontent.com/nlogozzo/NickvisionMoney/main/NickvisionMoney.Shared/Docs/html/{lang}/index.html"));
```
nlogozzo, line 192: Could you make the link: 
```csharp
$"https://htmlpreview.github.io/?https://github.com/nlogozzo/NickvisionMoney/blob/{_controller.AppInfo.Version}/NickvisionMoney.Shared/Docs/html/{lang}/index.html"
```

That way it uses the correct docs based on the version of the application and not the one from the main branch. Since, if a user hasn't updated the app and we change the docs on main, they will get docs for features that don't exist in their version.

nlogozzo (changes requested):

NickvisionMoney.WinUI/Views/PreferencesPage.xaml.cs
fsobolev: Done

PR #217, opened 2023-01-16
Add more metainfo tags and Spanish translation

oscfdezdz:


NickvisionMoney.GNOME/org.nickvision.money.metainfo.xml
```diff
@@ -5,10 +5,14 @@
...
   <name>Denaro</name>
   <summary>A personal finance manager</summary>
+  <summary xml:lang="es">Un gestor de finanzas personales</summary>
   <description>
     <p>
       A personal finance manager
     </p>
+    <p xml:lang="es">
+      Un gestor de finanzas personales
+    </p>
```
fsobolev, line 15: [translate_meta.py](https://github.com/nlogozzo/NickvisionMoney/blob/main/NickvisionMoney.GNOME/translate_meta.py) adds strings here automatically from Strings.*.resx during the build, so these should be removed. Also please remove `<summary>`, I will update the script so it will add it automatically too.

fsobolev (changes requested):

fsobolev (approved):

PR #223, opened 2023-01-17
Fix Dialog CPU Usage

### LINKED ISSUE #221: Fix modal dialog - 100% CPU core usage
The problem:
Dialogs are kept open in a loop using `g_main_context_iteration` that allows us to validate the data provided by the user in the dialog once it is closed. However, this causes a CPU core to have a usage of 100%, as the loop is continuously running until the dialog closes.

The solution:
Validate data as soon as the `Adw.EntryRow`s and other controls are updated, instead of when the dialog is closed. This will remove the need for a `g_main_context_iteration` loop

nlogozzo:
Introduces a new API for dialogs that removes the need to keep dialogs running with a loop (which is what caused the 100% CPU usage)

This new API also allows for the validation of dialog input as soon as input is changed.

Closes #221

### DIFF
--- a/NickvisionMoney.Shared/Models/AccountMetadata.cs
+++ b/NickvisionMoney.Shared/Models/AccountMetadata.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Data.Sqlite;
+using System;
 using System.IO;
 
 namespace NickvisionMoney.Shared.Models;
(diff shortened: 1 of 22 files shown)


NickvisionMoney.Shared/Models/AccountMetadata.cs
```diff
@@ -82,6 +83,29 @@ internal AccountMetadata(string name, AccountType accountType)
         SortTransactionsBy = SortBy.Id;
     }
 
+    /// <summary>
+    /// Clones the transaction
+    /// </summary>
+    /// <returns>A new transaction</returns>
```
fsobolev, line 89: Wrong description

fsobolev (approved):

PR #257, opened 2023-01-23
UI improvements

### LINKED ISSUE #201: Add action to open new window
Currently in GNOME there are 2 ways to open a new window when the app is running: either using terminal or by opening a file from file manager. Adding a menu option to open a new window will be useful in case a user wants to see 2 or more accounts at the same time on screen without using tabs.

### LINKED ISSUE #248: add desktop categories and keywords
hi
# Categories
Currently application does not specify categories so desktop application menu setting this in Others tab.

Screenshot from 2023-01-19 08-06-52

## my suggestion
i added this values for categories:
`Categories=Office;Finance;`
https://specifications.freedesktop.org/menu-spec/latest/apas02.html

Screenshot from 2023-01-19 08-08-34

# Keywords
i think adding some keywords might be helpful for finding app with desktop search.
like money keyword for now gnome gives to me this 💰.
> finance keyword works because of application comment.

fsobolev:
* "New window" action (closes #201)
* Hide "Transactions" label in account view in compact mode, because depending on localization it can be ellipsized almost entirely, which doesn't look great
* Matrix chat link in about dialog
* Keywords in desktop file (closes #248)

### DIFF
--- a/NickvisionMoney.GNOME/Views/MainWindow.cs
+++ b/NickvisionMoney.GNOME/Views/MainWindow.cs
@@ -37,7 +37,7 @@ public partial class MainWindow : Adw.ApplicationWindow
     [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
     private static partial void gtk_show_uri(nint parent, string uri, uint timestamp);
 
-    private bool _compactMode;
+    public bool CompactMode;
     private readonly MainWindowController _controller;
     private readonly Adw.Application _application;
     private readonly Gtk.Box _mainBox;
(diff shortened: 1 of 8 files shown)


NickvisionMoney.GNOME/Views/MainWindow.cs
```diff
@@ -36,7 +37,7 @@ public partial class MainWindow : Adw.ApplicationWindow
     [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
     private static partial void gtk_show_uri(nint parent, string uri, uint timestamp);
 
-    private bool _compactMode;
+    public bool CompactMode;
```
nlogozzo, line 40: This should be a property. 
```csharp
public bool CompactMode { get; private set; }
```

nlogozzo (changes requested):

NickvisionMoney.GNOME/Views/MainWindow.cs
fsobolev: done

nlogozzo (approved):

Issue #274, opened 2023-02-01
Untranslated interface


Mejans:
Hello!
I've just installed the software via flatpak. But the interface is not translated into my language, whereas I did the job on weblate.

fsobolev:
Please provide additional information:
1. The app version
2. The output of `locale` command

Mejans:
Hello!
1. It's version 2023.1.1
2. The output of locale:
`LANG=oc_FR.UTF-8
LANGUAGE=oc:fr_FR:en_GB:en
LC_CTYPE="oc_FR.UTF-8"
LC_NUMERIC=oc_FR.UTF-8
LC_TIME=oc_FR.UTF-8
LC_COLLATE="oc_FR.UTF-8"
LC_MONETARY=oc_FR.UTF-8
LC_MESSAGES="oc_FR.UTF-8"
LC_PAPER=oc_FR.UTF-8
LC_NAME=oc_FR.UTF-8
LC_ADDRESS=oc_FR.UTF-8
LC_TELEPHONE=oc_FR.UTF-8
LC_MEASUREMENT=oc_FR.UTF-8
LC_IDENTIFICATION=oc_FR.UTF-8
LC_ALL=
`

fsobolev:
@nlogozzo
I forced localizer to use needed locale:
```
_resourceSet = _resourceManager.GetResourceSet(new CultureInfo("oc-FR"), true, true)!;
```
But this didn't help, interface is in English. Renaming the resx file to `Strings.oc-FR.resx` didn't help too. Any ideas?

Mejans:
You mean there is no fallback oc_FR to only "oc"?

fsobolev:
This seems to be a dotnet bug. For some reason, it doesn't build a required file for Occitan translation when building under Linux. We found some similar old issues, but they are fixed, so it's something different that hopefully will be fixed. Until then, the workaround is to build the file on Windows and have it in repository, then it will be copied during the build using flatpak manifest. So don't worry @Mejans, your translation will appear in all next versions. :)

![](https://i.imgur.com/Boud1xT.png)

nlogozzo:
#276

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.2.0-rc1-f1

Mejans:
Just landed in my computer :D Many thanks!

PR #276, opened 2023-02-01
GNOME - Workaround to build Occitan translation

fsobolev:
Related: #274 
I'm not sure if the issue should be closed until properly fixed.


nlogozzo:
I think we can close #274 with this.

PR #302, opened 2023-02-10
Fix things mentioned in GNOME Circle review

fsobolev:
https://gitlab.gnome.org/Teams/Circle/-/issues/143#note_1661740

Things that were not fixed:
1. > Buttons and especially dialogs should never have "OK" as a label, always use something specific to the action, e.g. "Add" in this case.

    Requires translations change

2. > The sidebar calendar doesn't seem to be keyboard-accessible, I can't tab into the actual calendar grid

    As I mentioned in the Circle issue, it is actually keyboard-accessible, but in a strange way. I don't think we can fix it without replacing Gtk.Calendar with a custom widget (which I think is a good idea since I'm not absolutely happy with how Gtk.Calendar looks even with our CSS).

Also I improved metadata file and replaced the link to Peek flatpak in CONTRIBUTION.md with Kooha, because it works good enough on Wayland and is an easy to use video recorder designed for GNOME.


NickvisionMoney.GNOME/Controls/TransactionRow.cs
```diff
@@ -130,17 +143,17 @@ public bool IsSmall
...
+                _iconCompact.SetVisible(true);
             }
             else
             {
-                _row.RemoveCssClass("row-small");
                 _boxSuffix.SetOrientation(Gtk.Orientation.Horizontal);
                 _boxSuffix.SetMarginTop(0);
                 _btnId.SetVisible(true);
+                _iconCompact.SetVisible(false);
             }
```
DaPigGuy, line 157:
```c#
_btnId.SetVisible(!_isSmall);
_iconCompact.SetVisible(_isSmall);
```

Minor cleanup

fsobolev: thanks :)

nlogozzo (approved):

PR #309, opened 2023-02-16
GNOME - Improve styling of amount labels

fsobolev:
Amount labels are not inside buttons, like it was before the latest release. They don't use `accent`, `success` and `error` CSS classes anymore, only existing custom classes (`denaro-income` etc; overriding works like before). These classes now set font weight to bold and use named colors from Adwaita palette, depending on current style (`red_1` for dark style and `red_5` for light style, same for other colors).

| Light | Dark |
| --- | --- |
|  |  |

This PR should finally resolve the contrast issue raised in Circle review.


nlogozzo:
Ah definitely much sharper! Thanks :)

nlogozzo (approved):

PR #311, opened 2023-02-17
Circle Fixes

DaPigGuy:
Don't seem to have access to push to main branch, so here's a PR. Marking as a draft until all issues are handled.

- [x] Buttons and especially dialogs should never have "OK" as a label, always use something specific to the action
- [ ] The welcome screen layout has a lot going on (illustration, recent accounts, pill buttons, description)
  - **Pig:** Not sure what to do here
- [x] In large text mode, the description on groups can break pretty badly.
image
- [x] The avatar thing in the top right of the account settings looks a bit odd and doesn't really add anything.
- [x] The preferences window doesn't close with ESC
- [x] The welcome screen has some elements that aren't activatable but get keyboard focus

https://gitlab.gnome.org/Teams/Circle/-/issues/143#note_1661740

### DIFF
--- a/NickvisionMoney.GNOME/Controls/GroupRow.cs
+++ b/NickvisionMoney.GNOME/Controls/GroupRow.cs
@@ -17,7 +17,8 @@ public partial class GroupRow : Adw.ActionRow, IGroupRowControl
     private readonly Gtk.Label _lblAmount;
     private readonly Gtk.Button _btnEdit;
     private readonly Gtk.Button _btnDelete;
-    private readonly Gtk.Box _box;
+    private readonly Gtk.Box _boxButtons;
+    private readonly Gtk.FlowBox _flowBox;
 
     /// <summary>
     /// The Id of the Group the row represents
@@ -51,6 +52,8 @@ public partial class GroupRow : Adw.ActionRow, IGroupRowControl
         _cultureDate = cultureDate;
         //Row Settings
         SetUseMarkup(false);
+        SetTitleLines(1);
+        SetSubtitleLines(2);
         //Filter Checkbox
         _chkFilter = Gtk.CheckButton.New();
         _chkFilter.SetValign(Gtk.Align.Center);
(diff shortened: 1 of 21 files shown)


fsobolev:
> The welcome screen layout has a lot going on (illustration, recent accounts, pill buttons, description)
> * **Pig:** Not sure what to do here

My opinion, that I previously said to Nick privately:

> that's definitely a pattern that is not used in other GNOME apps, but the fact that it wasn't used doesn't mean it can't be.  I like this greeting too, so until somebody will prove that it contradicts GNOME HIG I would insist on keeping it.

The greeting fills empty space well and gives warming welcome to users.
But with avatar it's different, it really doesn't have any purpose, just a toy thing that can be easily dropped (although we don't have to drop it in WinUI).

> * In large text mode, the description on groups can break pretty badly.

I suppose we can detect somehow when large text mode is enabled and change the layout.
[GTK Accessibility docs](https://docs.gtk.org/gtk4/section-accessibility.html)

DaPigGuy:
I've dropped the avatar on GNOME. @nlogozzo can figure out if it should stay on Windows.

Do we have any other ideas for the start page?

fsobolev:
Since we don't need custom heading in AccountSettingsDialog anymore I changed it.

Снимок экрана от 2023-02-17 10-43-32

nlogozzo:
> Do we have any other ideas for the start page?

I mean I really feel as that the start page isn't that bad. Like @fsobolev said, I love the welcome message and don't want that to change. We put a lot of work making sure the start page still looks good on small screens so I don't know what else to change 🤷‍♂️

nlogozzo:
I just want to change the dialog labels in WinUI as well, then I'll merge :)

nlogozzo (approved):

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.2.1

PR #329, opened 2023-02-25
Circle Fixes 4

DaPigGuy:
Yet another sequel

image
image
image

### DIFF
--- a/NickvisionMoney.GNOME/Resources/org.nickvision.money.css
+++ b/NickvisionMoney.GNOME/Resources/org.nickvision.money.css
@@ -20,8 +20,13 @@
 .account-tag {
   border-radius: 20px;
   font-size: 13px;
-  padding-left: 5px;
-  padding-right: 5px;
+  padding-left: 13px;
+  padding-right: 13px;
+}
+
+.greeting-title {
+  font-size: 32px;
+  font-weight: 900;
 }
 
 /* Calendars colors */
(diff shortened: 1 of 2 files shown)


DaPigGuy:
The account type tags are no longer the same width w/o the width request. Not sure if that's ideal but it would already affect languages where the text extended too far anyway.

NickvisionMoney.GNOME/Views/MainWindow.cs
```diff
@@ -233,6 +235,7 @@ public MainWindow(MainWindowController controller, Adw.Application application)
         _btnNewAccount.AddCssClass("pill");
         _btnNewAccount.AddCssClass("suggested-action");
         _btnNewAccount.SetDetailedActionName("win.newAccount");
+        _btnNewAccount.SetMarginBottom(12);
```
fsobolev, line 238: This will add space between buttons when they will be vertical. It's better to add margin to flowbox.

NickvisionMoney.GNOME/Resources/org.nickvision.money.css
```diff
@@ -20,8 +20,13 @@
 .account-tag {
   border-radius: 20px;
   font-size: 13px;
-  padding-left: 5px;
-  padding-right: 5px;
+  padding-left: 13px;
+  padding-right: 13px;
```
fsobolev, line 24: As I understood Tobias asked to decrease paddings to make it look similar to Builder, not increase it.

fsobolev (changes requested):

NickvisionMoney.GNOME/Resources/org.nickvision.money.css
DaPigGuy: The high amount of "padding" was the width request set to 100

fsobolev: Ah, okay. I looked at Builder's tags and they have 12px paddings and ~12.2px font-size. This is very close, so it's fine to leave as it is now. Or you can change, up to you.

DaPigGuy: 13px looks fine :)

fsobolev (approved):

PR #355, opened 2023-03-10
More custom currency settings

### LINKED ISSUE #318: Add more decimal places
Hi,

Is it possible to add more decimal places in amounts to support cryptocurrencies. It sounds like currently there is only 2 decimal places. So an amount of `0.001` is rounded to `0.00`. But for money like Bitcoin, `0.001 ₿` is around `20 $` so decimals are not negligible.

Thanks for your work anyway!

fsobolev:
I hoped to finish everything for GNOME in one day, but I underestimated the amount of work required to implement it 😄 Still, here's the draft.

This PR adds 3 new settings to custom currency:
1. Decimal separator: `.`, `,` or custom
2. Group separator: `.`, `,`, `'`, none or custom
3. Decimal digits number: 2-6 or unlimited (closes #318)

TODO:
- [x] Fix database to accept empty group separator
    I didn't think about it when writing database related code, group separator is handled the same way as decimal separator (that can't be empty). Nick, I will be grateful if you can work on this, because I'm afraid to do it wrong, this stuff is a bit complicated to me 😅 
- [x] Use Gtk.Stack instead of expander row for custom currency settings
    Similar to what we have with password settings in WinUI. The dialog on the screenshot above doesn't actually fit on my screen (1366x768).
- [x] Save settings (currently only validation is implemented)
- [x] WinUI port
- [ ] Update docs

Some notes:
* Max length for custom separators is set to 2 to allow emojis
* "Unlimited" decimal digits are actually 99, the biggest acceptable number
* Decimal and group separator can't be the same. But I'm not sure how to mark it in the dialog in a short and nice way. Currently rows get `error` CSS class, but labels are not changed.
* I created separate strings for "Other" separators. The reason is that we can't be sure that in all languages this word will be the same for both separators
    <details>
        <summary>Explanation</summary>
        For example, in Russian words have <a href="https://en.wikipedia.org/wiki/Grammatical_gender">grammatical gender</a>. Depending on it, "other" can be translated as "другой", "другая" or "другие". For separators it's not the case ("другой" for both), but I can imagine that maybe not for this particular reason but because of some other rules some languages will require to translate it differently for each separator.
    </details>


NickvisionMoney.GNOME/Blueprints/account_settings_dialog.blp
```diff
@@ -81,6 +81,54 @@ Adw.MessageDialog _root {
...
           }
         }
+
+        Adw.ActionRow _customDecimalSeparatorRow {
+          title: _("CustomCurrencyDecimalSeparator.Field");
+
+          [suffix]
+          Gtk.DropDown _customDecimalSeparatorDropDown {
+            valign: center;
+            show-arrow: false;
```
nlogozzo, line 91: Why'd you hide the arrows for these properties? I get the design choice, but I feel like for an accessibility stand point it might make it easier for the user to know it is a drop down with the arrow there. Unless there is some GNOME HIG I don't know about that says hidden the arrow is ok :)

nlogozzo (commented):
> Nick, I will be grateful if you can work on this, because I'm afraid to do it wrong, this stuff is a bit complicated to me 😅

Got it :) . I believe I can change the column in SQL to accept empty strings but I have to make sure it doesn't break existing files...

NickvisionMoney.GNOME/Blueprints/account_settings_dialog.blp
fsobolev: The only reason is that I wanted to save space for when "Other" is selected and there are row label, dropdown and entry in one line in pretty narrow dialog. I will think how to change it to look better and keep arrows. Probably I will change entries placeholder to something short so it will not require much space, it's not possible to enter more than 2 characters so it's kinda waste currently.

fsobolev:
> I believe I can change the column in SQL to accept empty strings but I have to make sure it doesn't break existing files...

If accepting empty string is problematic, we can save literally `"empty"` string in database. When this string is detected, empty string is assigned in metadata instead of the actual value.

fsobolev:
I will work on the rest tomorrow, but currently settings are saved and used!

nlogozzo:
Thank you :) rest up!!

nlogozzo:
Ok, custom separators are implemented in WinUI. But I would like to tweak the design similar to how there is a separate password page in WinUI and clean up some of the backend.

I'd like to get this done tonight and then release beta 2 with this feature and all the fixes for pdf, qif, and ofx.

nlogozzo:
image
image

Here's the new currency card design similar to that of the password card :)

nlogozzo:
@fsobolev All that's left is for you to make whatever design tweaks needed on GNOME and update the docs and we are good to go!

nlogozzo:
Also implemented all the code cleanups I wanted too :)

fsobolev:
I have an idea for a smarter decimal separator insert. But idk if it's possible to conveniently do the same in WinUI, so I don't commit changes for now, because if it can't work the same on both platforms, then I think it's not worth it.
In GNOME instead of:
```
row.SetText(row.GetText() + _controller.CultureForNumberString.NumberFormat.CurrencyDecimalSeparator);
row.SetPosition(row.GetText().Length);
```
do this:
```
if (!row.GetText().Contains(_controller.CultureForNumberString.NumberFormat.CurrencyDecimalSeparator))
{
    var position = row.GetPosition();
    row.SetText(row.GetText().Insert(position, _controller.CultureForNumberString.NumberFormat.CurrencyDecimalSeparator));
    row.SetPosition(position + Math.Min(_controller.CultureForNumberString.NumberFormat.CurrencyDecimalSeparator.Length, 2));
}
```

So the separator will be inserted in cursor position, not always in the end of entry. `if (!row.GetText().Contains...` and `Math.Min(...` are required to make it correctly work with emojis otherwise position will not be calculated properly if the separator is 2-character emoji. And anyway it doesn't make sense to insert more than 1 decimal separator.

nlogozzo:
This will be possible to achieve on WinUI. Go ahead and commit it and I will port it over before merge :)

nlogozzo:
WinUI implementation done :)

fsobolev:
In latest commit I added more currency checks. Now it's not possible to set a number as currency symbol and also symbol can't contains one of separators. Not only it makes sense, but it prevents breaking amount formatting:
```
        if (!showCurrencySymbol)
        {
            result = result.Remove(result.IndexOf(culture.NumberFormat.CurrencySymbol), culture.NumberFormat.CurrencySymbol.Length);
        }
```

If because of system locale the culture sets currency symbol to be show on right, and if symbol would be a number or one of separators, this will remove the first matching number/separator instead of currency symbol.
I also added changes to WinUI but didn't test it.

nlogozzo:
From first look, WinUI looks good to me. Will run the code when I get on my pc.

nlogozzo:
Ok, it all looks good!

What's left to do before merge?

fsobolev:
I'm working on changing the UI... Currently stuck with making stack work properly, but don't worry, I will finish it today.
Then the only thing that is left is documentation but this can be done later.

nlogozzo:
> I'm working on changing the UI... Currently stuck with making stack work properly, but don't worry, I will finish it today.

Thank you :)

> Then the only thing that is left is documentation but this can be done later

As long as we do it before release that's fine. Beta will be OK without it.

fsobolev:
Okay, now account settings dialog is beautiful and very similar to the one in WinUI.

| Main | Currency | Password |
| --- | --- | --- |
|  |  |  |

What you can see on screenshots:
1. If something on currency or password page is invalid, a corresponding row on main page gets error class telling a user that pressing "back" on other pages doesn't reset values and errors should be fixed.
2. Currency separators entries are now smaller and don't have placeholders, but they get focused immediately after selecting "Other" in drop down so it's clear that it's an entry and not an empty button or something else.
3. Since password changing functionality is not on main page anymore, I decided that it would be better to mark matching password errors similar to WinUI.

fsobolev:
Tbh I don't like the empty space next to "back" buttons, but I don't want to completely change showing errors to be everywhere like for password matching... Let me try hiding the dialog title when we're not on main page.

nlogozzo:
For the main page...instead of it saying "Use Custom Currency" could it just say "Currency" like WinUI, since you have to turn on the switch to use custom currency otherwise it just shows the reported system currency.

nlogozzo:
> Tbh I don't like the empty space next to "back" buttons, but I don't want to completely change showing errors to be everywhere like for password matching... Let me try hiding the dialog title when we're not on main page.

TBH I think it likes fine haha but work your magic if you'd like :)

fsobolev:
Hmmm, idk... Probably better not do this

image

fsobolev:
> since you have to turn on the switch to use custom currency otherwise it just shows the reported system currency.

When the switch is off the preference group containing all settings gets insensitive but not hidden, I think it's better to avoid even more dialog size changing.

nlogozzo:
> > since you have to turn on the switch to use custom currency otherwise it just shows the reported system currency.
> 
> When the switch is off the preference group containing all settings gets insensitive but not hidden, I think it's better to avoid even more dialog size changing.

I was justifying for changing the label on the main page from "Use Custom Currency" --> "Currency" so all good

nlogozzo (approved):

nlogozzo:
Looks good to me! Will merge and release beta 2 when I get home in about an gour from work :)

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.3.0-beta2

PR #356, opened 2023-03-10
GNOME - Fix files opening (use Gtk.FileDialog; open files on start)

fsobolev:
Hello a ton of marshalling, long time no see 😄 

### The problem

When running with GTK 4.9+, `Gtk.FileChooserNative` becomes unusable (missing labels and icons - [screenshot](https://user-images.githubusercontent.com/117388856/224332396-2031cf63-67ad-423b-8e55-ac8b75831370.png) - trying to blindly open a file crashes the app).
I'm not entirely sure what causes the problem, but I guess it's somehow related to gir.core because I didn't find any other app other than Denaro and TC that has this bug. But I think gir.core devs have more important stuff than fixing deprecated things, so I didn't report it.

### Temporary solution

`Gtk.FileChooserNative` and `Gtk.FileChooserDialog` are deprecated in favor of `Gtk.FileDialog` added in GTK 4.9.1. In this PR Denaro uses this new widget when running with supported GTK version.
I tested this PR in Fedora 37 (GTK 4.8.3), Fedora 38 Beta (GTK 4.9) and in Arch distrobox (GTK 4.10).

### Long-term solution

When GNOME 44 Platform runtime will be released with GTK 4.10 I suggest to use it and completely remove `Gtk.FileChooserNative`. This will make 4.10 minimum required version of GTK, but I think people who use distros with old libraries should use Flatpak anyway, so this will not be a problem. And when gir.core will support GTK 4.10 we'll get rid of marshalling.
Because the code will be changed, I don't think we should copy everything to Application now. But after accepting this PR I will open a new PR for TC.

### A new small issue

`Gtk.FileDialog` when used to open a file has a checkbox to open a file in read-only mode that can't be hidden. This doesn't break Denaro, because we only use resulting `Gio.File` to get the path and then Denaro opens the file read-write anyway. But this can be misleading for users, so maybe we should consider adding read-only mode to Denaro in the future.

**UPD**: Also fixed files opening on start.

### DIFF
--- a/NickvisionMoney.GNOME/Views/MainWindow.cs
+++ b/NickvisionMoney.GNOME/Views/MainWindow.cs
@@ -81,6 +81,9 @@ public partial class MainWindow : Adw.ApplicationWindow
     private readonly Gio.SimpleAction _actOpenAccount;
     private readonly Gio.SimpleAction _actCloseAccount;
 
+    private GAsyncReadyCallback _saveCallback { get; set; }
+    private GAsyncReadyCallback _openCallback { get; set; }
+
     public bool CompactMode { get; private set; }
 
     /// <summary>
@@ -311,10 +314,9 @@ public partial class MainWindow : Adw.ApplicationWindow
             var filters = Gio.ListStore.New(Gtk.FileFilter.GetGType());
             filters.Append(filter);
             gtk_file_dialog_set_filters(saveFileDialog, filters.Handle);
-            var fileHandle = IntPtr.Zero;
-            GAsyncReadyCallback callback = async (source, res, data) =>
+            _saveCallback = async (source, res, data) =>
             {
-                fileHandle = gtk_file_dialog_save_finish(saveFileDialog, res, IntPtr.Zero);
+                var fileHandle = gtk_file_dialog_save_finish(saveFileDialog, res, IntPtr.Zero);
                 if (fileHandle != IntPtr.Zero)
                 {
                     var path = g_file_get_path(fileHandle);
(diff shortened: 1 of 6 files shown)


nlogozzo:
Will test tonight when I get home on WSL, make sure it's ok then merge :)

nlogozzo:
> Will test tonight when I get home on WSL

LGTM. If you could just test the latest commits to see if that opening file on startup is fixed.

NickvisionMoney.GNOME/Views/MainWindow.cs
```diff
@@ -278,32 +301,65 @@ private async void OnWindowShow(Gtk.Widget sender, EventArgs e)
...
+        if (Gtk.Functions.GetMinorVersion() >= 9)
         {
-            if (e.ResponseId == (int)Gtk.ResponseType.Accept)
+            var saveFileDialog = gtk_file_dialog_new();
+            gtk_file_dialog_set_title(saveFileDialog, _controller.Localizer["NewAccount"]);
+            var filters = Gio.ListStore.New(Gtk.FileFilter.GetGType());
+            filters.Append(filter);
+            gtk_file_dialog_set_filters(saveFileDialog, filters.Handle);
+            var fileHandle = IntPtr.Zero;
+            GAsyncReadyCallback callback = async (source, res, data) =>
```
nlogozzo, line 315: Shouldn't these be saved as instance variables to avoid that error "calling delegate on garbage collection object"...

nlogozzo (approved):

nlogozzo:
https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.3.0-beta1

PR #365, opened 2023-03-14
Fix PDF Header Columns Spacing

### LINKED ISSUE #360: Change rows width in generated PDF
Not sure about other rows, but at least Id row should be changed to fit translations that contain more than 3 characters.

nlogozzo:
Fixes #360

### DIFF
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -1593,7 +1593,7 @@ public class Account : IDisposable
                                 }
                                 tbl.Cell().Background(hex).Text(pair.Value.Id.ToString());
                                 tbl.Cell().Background(hex).Text(pair.Value.Date.ToString("d", cultureDate));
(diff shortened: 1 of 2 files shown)


fsobolev:
Doesn't work good

nlogozzo:
ok, try now

fsobolev:
All column headers now looks good for me. "Еженедельно" ("Weekly") is badly split (you can see it on the screenshot above), but I don't think we can do anything about it: even if the column will be a bit wider to fir russian translation. in an another language there can be a few more characters that will cause similar result. So we can't guarantee everything to always fit correctly anyway.
It's good you decided to work on this and OFX import at the same time 😄 I imported an OFX file containing a transaction with whitespaces in the end of the description, and this caused PDF export to fail. `Trim()` fixes it (https://github.com/nlogozzo/NickvisionMoney/pull/365/commits/ac4400607f1766536bff67387fe67d68773c6ccb).

fsobolev (approved):

PR #371, opened 2023-03-14
Fix QIF loading with non-English locale

### LINKED ISSUE #368: QIF import broken
I tried to import data from an other app, HomeBank, as well as the example file given in #265 and the toaster says "0 transactions imported".

The format of the date in both files is correct (US date: MM/DD/YYYY).

I'm using the latest version at the time of writing, which is [`2023.2.2`](https://github.com/nlogozzo/NickvisionMoney/releases/tag/2023.2.2).

A QIF file that is broken: [test.zip](https://github.com/nlogozzo/NickvisionMoney/files/10970843/test.zip)

fsobolev:
`QifDocument.Load` fails if the current culture is not en-US. Function call is now wrapped in `Task.Run` to be executed in separated thread where culture is changed without affecting the rest of the app.
Closes #368

### DIFF
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -1323,10 +1323,10 @@ public class Account : IDisposable
         QifDocument? qif = null;
         try
         {
-            qif = Task.Run(() => {
-                Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
-                return QifDocument.Load(File.OpenRead(path));
-            }).Result;
+            var oldCulture = Thread.CurrentThread.CurrentCulture;
+            Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
+            qif = QifDocument.Load(File.OpenRead(path));
+            Thread.CurrentThread.CurrentCulture = oldCulture;
         }
         catch
         {
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -14,6 +14,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace NickvisionMoney.Shared.Models;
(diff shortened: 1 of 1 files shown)


NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -1322,7 +1323,10 @@ private async Task<List<uint>> ImportFromQIFAsync(string path, string rgba)
         QifDocument? qif = null;
         try
         {
-            qif = QifDocument.Load(File.OpenRead(path));
+            qif = Task.Run(() => {
```
nlogozzo, line 1326: Task.Run is awaitable so you should be able to make it 
```csharp
qif = await Task.Run(() =>
{
    Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
    return QifDocument.Load(File.OpenRead(path));
});
```
and remove the need for `.Result`

nlogozzo (approved):

NickvisionMoney.Shared/Models/Account.cs
nlogozzo: And actually, `Task.Run` and any `Task` for that matter isn't truly guaranteed to run in a separate thread. 
So it's probably best to just make it
```csharp
var oldCulture = Thread.CurrentThread.CurrentCulture;
Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
qif = QifDocument.Load(File.OpenRead(path));
Thread.CurrentThread.CurrentCulture = oldCulture;
```
within the try block itself.

nlogozzo (changes requested):
Didn't mean to approve...that was accident :)

NickvisionMoney.Shared/Models/Account.cs
fsobolev: > And actually, `Task.Run` and any `Task` for that matter isn't truly guaranteed to run in a separate thread.

Ouch, I though it is, that's why I preferred to use it instead of changing the culture in current thread like you said. But I guess there's no better option

nlogozzo: > But I guess there's no better option

Yeah...i mean even if I were to send a patch the the library itself, i'd pretty much be the same code. Change the culture to english before, load the qif, change it back to previous. So...

nlogozzo (approved):

PR #383, opened 2023-03-16
Add CSV Backup Folder Preference Option

### LINKED ISSUE #334: [Feature Request] Export automatically after making a change in my account
Basically, I would appreciate if possible the ability to automatically export a copy of my account after making a change (transaction) to a previously defined directory. 

I mainly request this to have a backup on my Nextcloud server, but it could also serve if I wanted to synchronize Denaro with any mobile apps that support CSV.

nlogozzo:
This PR implements a `CSV Backup Folder` option in preferences.

Closes #334

### DIFF
--- a/NickvisionMoney.Shared/Resources/Strings.resx
+++ b/NickvisionMoney.Shared/Resources/Strings.resx
@@ -1023,6 +1023,34 @@
 		<value>Period and Comma</value>
 	</data>
 
+	<data name="Backup" xml:space="preserve">
+		<value>Backup</value>
+	</data>
+
+	<data name="BackupDescription" xml:space="preserve">
+		<value>Customize backup locations for Denaro accounts.</value>
+	</data>
+
+	<data name="BackupFolder" xml:space="preserve">
+		<value>CSV Backup Folder</value>
+	</data>
+
+	<data name="BackupFolderDescription" xml:space="preserve">
+		<value>Accounts will automatically export as CSV files to this folder when a change is made.</value>
+	</data>
+
+	<data name="NoBackupFolder" xml:space="preserve">
+		<value>No Backup Folder</value>
+	</data>
+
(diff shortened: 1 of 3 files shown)


fsobolev:
Suggestion: CSV export can take some time (usually small, but still), so we should prevent the app closing while backup creation is in process. Maybe a dialog:
> CSV Backup is in process
The app will be closed when backup will be finished.
[Cancel]

Or we could not notify a user but simply let the app to continue running in the background.

nlogozzo:
The `ExportToCSV` function is not async, meaning the UI will freeze for a few milliseconds while the file is exported and the user shouldn't even notice it really. Because the UI freezes, the window can't be closed until it is done.

I'll do some testing of the feature on big accounts, 1000+ transactions, and if the export is noticeably long, I'll make the function async. If it's async and the window is closed, the export will continue running in the background until the function finishes and then the whole application will close.

fsobolev:
I need to leave home for a couple of hours but then I can finish it.

nlogozzo:
> I need to leave home for a couple of hours but then I can finish it.

No worries if I have some free time before you I'll finish it up.

fsobolev:
@DaPigGuy Regarding the last commit (https://github.com/nlogozzo/NickvisionMoney/pull/383/commits/c68d14ea506cf81bbc8bd3033dcec48596526841), so you will know too:
> The problem was we used `Adw.PreferencesPage` wrong. It is intended to be used with other `Adw.Preferences*` widgets and doesn't require anything else: it has title, overlay etc.
But we used `Gtk.Box` as its child, placing headerbar and `Adw.PreferencesPage` here. And it basically turned PreferencedWindow to a simple Window

While everything looked good, PreferencesWindow with "overwritten" layout can't show toasts, although there's no error on `AddToast` call in this case.

PR #393, opened 2023-03-20
Dashboard Page & Group Colors

### LINKED ISSUE #162: Opened Accounts Overview Page
From #90:
> A cross account Dashboard (I should note that there should still be a way to separate accounts, just that it is also useful to have a summary of multiple ones)

Maybe we could add a separate tab page that just shows data for all accounts:
- Total amount
- Total income
- Total expense

But I wouldn't really know what else to add to this page. @RuboGubo could you give some suggestions?

### LINKED ISSUE #339: Group colors
>It would be cool if i can link a color to a group. For example i want, that if i add a expese/income with the group "food" the group sets automaticly the color to... any color i set. That would be cool

Requested by programmtube:matrix.teckids.org

We could add color row to group dialog, make groups checkboxes colored, and in transaction dialog add dropdown next to color button: it will have 2 options - "use unique color" and "use group color", the default value will be set depending on a setting (in preferences, I don't think it needs to be per account) and when no group is selected it will be disabled.
In CSV, a new column can be created for group color and for RGBA "g" or something else can be used to mark that this transaction uses group color.

### LINKED ISSUE #407: Remove scroll on "insert password" window
Looks like it shouldn't have scroll abilities there, because there's nothing to show below the password input

[Screencast from 2023-03-29 14-34-48.webm](https://user-images.githubusercontent.com/92032579/228447766-3c97ddfa-bdcb-4810-9844-693ef019b56c.webm)

nlogozzo:
Closes #162 
Closes #339 
Closes #407 

## TODO
- [X] Redesigned AccountViewPage for WinUI to better match fluent design
- [x] Group Colors
    - [x] Backend
    - [x] GNOME
    - [x] WinUI 
    - [x] All: TransactionDialog -> Change To Group Color When Group Selected
    - [x] Shared: Ungrouped color match config
- [x] DashboardViewController
- [x] GNOME Implementation
- [x] WinUI Implementation
- [x] Fix password dialog in GNOME


nlogozzo:
> Redesigned AccountViewPage for WinUI to better match fluent design

Previous WinUI design was something that was a directly design copy from GNOME. While the design works really well on GNOME, it did not on WinUI. It felt very congested and at times complicated to use, so I redid it! Furthermore, in keeping with WinUI design, I moved the tab view from the top of the application to a navigation sidebar.
Here is the new redesigned, WinUI design:

image
image
image
image
image

fsobolev:
"Ungrouped" group uses starting color (green), even if the default group color was changed in the settings, I think "Ungrouped" should always use user-defined default color.

nlogozzo:
> All: TransactionDialog -> Use Group Color / Use Unique Color

I feel like having this dropdown will create a complicated UX/UI design. What if, when the user selects a group from that drop-down, the color button will automatically change to that group's color, and then if the user wants they can ofc course click the button and change the color...and if the group is `Ungrouped` it will go back to `TransactionDefaultColor`

nlogozzo:
> > All: TransactionDialog -> Use Group Color / Use Unique Color
> 
> I feel like having this dropdown will create a complicated UX/UI design. What if, when the user selects a group from that drop-down, the color button will automatically change to that group's color, and then if the user wants they can ofc course click the button and change the color...and if the group is `Ungrouped` it will go back to `TransactionDefaultColor`

This is how I implemented it.

fsobolev:
> > All: TransactionDialog -> Use Group Color / Use Unique Color
> 
> I feel like having this dropdown will create a complicated UX/UI design. What if, when the user selects a group from that drop-down, the color button will automatically change to that group's color, and then if the user wants they can ofc course click the button and change the color...and if the group is `Ungrouped` it will go back to `TransactionDefaultColor`

I wanted to slightly change the dialog anyway, because I'm not happy with horizontally stacked rows, but for this feature I think we should change part of the dialog to something like this:

Снимок экрана от 2023-04-05 15-54-53

nlogozzo:
> I wanted to slightly change the dialog anyway, because I'm not happy with horizontally stacked rows, but for this feature I think we should change part of the dialog to something like this:

I mean I feel like adding this `Use Unique Color` switch makes it unintuitive. Cause the way I implemented it, if you select a group it's color become's the color button's color and if the user wants to use a unique color, they can just change the button like always. No need to turn an extra switch on or off.

fsobolev:
> I mean I feel like adding this `Use Unique Color` switch makes it unintuitive. Cause the way I implemented it, if you select a group it's color become's the color button's color and if the user wants to use a unique color, they can just change the button like always. No need to turn an extra switch on or off.

I think the opposite. The relation between color and group is not intuitive. Making color row and group row linked, with the label of color row that doesn't just say "color" but implies that it's related to a group, makes this relation obvious. The color button will change when a group is changed if the switch is off. If it's on, the color selected by user will remain even if the group was changed. And if a user wants to reset color to use the one provided by group, they will simply toggle the switch off.

nlogozzo:
I mean I see what your saying but I think the most common use case will be:
- User creates a new transaction
- User selects a group (the group's color now becomes the transaction's color)
- The user decides to have a unique color (simply changes the color button, your way they'd have to turn on the switch then select a color)
- Saves the transaction
If the user decides to change the group:
- When the user selects the new group, little drawback the color button changes to the new group's color
- The user may decide to leave this new color and leave it, or if they want to change it they can change it back again (your way, there will have to be no changing back)

Either way it's implemented there will have to be an extra click for certain use cases. An extra click your way if they want a unique color or an extra click my way if they want to keep that unique color when changing a group later on.
So I mean, I don't know either way it's fine to me.

If you want to implement it your way on GNOME, feel free :) . I think I'll leave it my way on WinUI because the dialog is already pretty full and I don't know how I would move it all around to include a new switch.

fsobolev:
> I mean I see what your saying but I think the most common use case will be:

With my proposal I wasn't trying to improve most common use cases or decrease the number of clicks, that's not an issue. What I wanted to achieve is to provide a UI that gives the user as much information as possible *before* a user touches something for the first time and see how things work.

> if they want to change it they can change it back again (your way, there will have to be no changing back)

Toggling the switch on and setting a desired color is the way to change back.

> the dialog is already pretty full and I don't know how I would move it all around to include a new switch.

Same problem in GNOME, I will try to figure out what to do with the dialog while keeping in mind that we'll also need a place for tags row in the future.

nlogozzo:
> > if they want to change it they can change it back again (your way, there will have to be no changing back)
> 
> Toggling the switch on and setting a desired color is the way to change back.

That's what I meant. If the toggle switch is already on there is no need to change the color, as it would already be unique and the group color has no effect. With my way, if a new group is selected and the color is unique, the group color will override and the user will have to change it back to the unique color. Sorry for any misunderstanding :)

> Same problem in GNOME, I will try to figure out what to do with the dialog while keeping in mind that we'll also need a place for tags row in the future.

👍 . I was thinking of maybe creating separate card pages similar to how we have in AccountSettingsDialog?

fsobolev:
[Kooha-2023-04-06-06-24-18.webm](https://user-images.githubusercontent.com/117388856/230264734-e4adbfcd-b444-44df-baae-bdf552f79fcd.webm)

`flat` class was removed from date buttons to make it clear that these are buttons, not labels (we used arrows before, but I think they look bad with new layout).

nlogozzo:
Looks great!

But I think text should say `Use unique color` instead of `Use specific color`...conveys the intent better

fsobolev:
> But I think text should say `Use unique color` instead of `Use specific color`...conveys the intent better

I thought that "unique color" could be misunderstood as "color that can be used for only this transaction and not for any other". But sure, if you say "unique" will be better, I will change it

fsobolev:
Снимок экрана от 2023-04-06 07-03-45

I decided to use the same trick we have with playlist in TC to make the dialog fully adaptive (it can't be resized by the user, I changed its size only to simulate small screen, if the window fits the screen there will be no scrollbar and borders).

nlogozzo:
WinUI done :)
On WinUI, if the `Ungrouped` group is selected I have the drop-down hidden and i just show the color button.
Maybe you can do the same on GNOME. Or if the `Ungrouped` group is selected maybe have the drop-down say, use default color or use unique color. Because when the `Ungrouped` group is selected, it will use the default transaction color and not the group color.

fsobolev:
> WinUI done :) On WinUI, if the `Ungrouped` group is selected I have the drop-down hidden and i just show the color button. Maybe you can do the same on GNOME. Or if the `Ungrouped` group is selected maybe have the drop-down say, use default color or use unique color. Because when the `Ungrouped` group is selected, it will use the default transaction color and not the group color.

What if a user wants to treat "Ungrouped" as regular group (and in fact it is a regular group, it also has a color) and assign some transactions to it also wishing for these transactions to use the group's color (= default group color)? I don't see a point in disallowing this.
But this is probably will be not very common workflow, using default transaction color would be preferable I think. And this is easy to change: `UseGroupColor` should be false by default (currently it's true).

nlogozzo:
> What if a user wants to treat "Ungrouped" as regular group (and in fact it is a regular group, it also has a color) and assign some transactions to it also wishing for these transactions to use the group's color (= default group color)?

At this point, it removes the purpose of Default Transaction Color since everything will use default group color then, unless the user specifies something different, hence why I was saying to change the wording to just use default color when the selected group is `Ungrouped`.

> But this is probably will be not very common workflow, using default transaction color would be preferable I think. And this is easy to change: UseGroupColor should be false by default (currently it's true).

If the group is set to `Ungrouped` it currently does uase the default transaction color as we have it where if the groupid is <= 0 (0 being ungrouped) use transaction color else group color. So having default `UseGroupColor` be false would have no effect. I think leaving it as true is fine.

fsobolev:
> At this point, it removes the purpose of Default Transaction Color

Not really. Default transaction color can be used for specific transactions no matter what the group is. It's kinda creating a "group" of transactions without actually changing their groups, just by marking them visually. And using default transaction color for this can save a few clicks.

> If the group is set to Ungrouped it currently does uase the default transaction color as we have it where if the groupid is <= 0 (0 being ungrouped) use transaction color else group color.

Does it? Well... Again, I don't see a point in disallowing using the color of "Ungrouped" group for transactions. I think this should be changed.

fsobolev:
About this commit: https://github.com/nlogozzo/NickvisionMoney/pull/393/commits/874836818a0454a445f8b77931772052eb13f169
No matter whether we want or not to have `UseGroupColor` set to true for new transactions, it should be false when creating an entry in database for backward-compatibility. Imagine what will happen when users will update from 2023.2.x to 2023.4.0 — they will have all transactions green. Because all groups will have green color and all transactions will follow it.

fsobolev:
> I don't see a point in disallowing this.

I found a point: it would be very rare workflow that doesn't worth increasing code complexity that would be required to pass default group color everywhere. So I made changed to match WinUI as you said. Color row is now not ComboRow, but ActionRow with DropDown that have a popover that looks a bit different than a popover of ComboRow (different position and arrow)... But I hope nobody will notice 😅

nlogozzo:
Looks good to me!

nlogozzo:
If nothing else, you can approve the PR and we can merge :)

fsobolev (approved):

nlogozzo:
Yay 🥳

PR #416, opened 2023-04-03
GNOME - Ensure GTK calls happen on main thread

### LINKED ISSUE #401: App Crashed after adding new transaction
Sometimes when i want to add new transaction the app is not responding and crashed. I've tried to run it from terminal and it showed this log before crashed

`malloc(): unaligned tcache chunk detected`

nlogozzo:
Closes #401 

Once and for all.

Thanks @fsobolev

### DIFF
--- a/NickvisionMoney.GNOME/Views/AccountView.cs
+++ b/NickvisionMoney.GNOME/Views/AccountView.cs
@@ -130,6 +130,7 @@ public partial class AccountView : Adw.Bin
     private readonly Gtk.Adjustment _transactionsScrollAdjustment;
     private readonly Gtk.ShortcutController _shortcutController;
     private readonly Action<string> _updateSubtitle;
+    private readonly GSourceFunc _startSpinner;
     private readonly GSourceFunc _stopSpinner;
     private readonly GSourceFunc _accountTransactionsChangedCallback;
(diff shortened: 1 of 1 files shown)


DaPigGuy:
Just noticed there are 11 instances of the same 6 lines copied for starting the spinner, may want to move that to its own method

nlogozzo:
@fsobolev maybe make start spinner the same way with the callback you did for stop spinner and "g invoke" it just to be safe

nlogozzo:
> Edit: You can further reduce the code duplication by having that method take in a Task, then:
Start spinner
Await Task
Stop spinner

i will be reduced to 3 lines

```
g_main_context_invoke(startSpinner)
await task
g_main_context_invoke(stopSpinner)
```

no need to make it it's own method for the 3 lines, especially if each task is different, we're just passing around the object for no reason (waste of memory)

nlogozzo:
> @fsobolev maybe make start spinner the same way with the callback you did for stop spinner and "g invoke" it just to be safe

@Krindog7337 confirmed no crashes .🎉
Could you do this and then we can merge?

nlogozzo:
>  maybe make start spinner the same way with the callback you did for stop spinner and "g invoke" it just to be safe

done.

fsobolev (approved):

PR #419, opened 2023-04-07
Better automatic backup

fsobolev:
1. Do not create backups for encrypted accounts, this makes encryption pointless
2. Recreate target folder while exporting (in case it was removed by user for example)


NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -1430,6 +1431,7 @@ public bool ExportToPDF(string path, string? password)
     {
         try
         {
+            Directory.CreateDirectory(System.IO.Path.GetDirectoryName(path));
```
nlogozzo, line 1434: Should this also be added to ExportToCSV Function?

nlogozzo (changes requested):

NickvisionMoney.Shared/Models/Account.cs
fsobolev: It is added (line 1414)

nlogozzo (approved):

nlogozzo:
Ah missed it 😅

PR #430, opened 2023-04-09
Group Color Fixes

### LINKED ISSUE #429: Transaction Using Group Color Doesn't Update On Group Color Change
From Matrix Room:
> If I create a transaktion and the transaction is every month (and I do this in the settings of the transaction) and then use the group color. After I create the transaction, I edit the group color. And then, the color isn't automatically updated and I have to edit the transaction, do nothing, save. And then the color is updated.

nlogozzo:
This PR fixes an issue where transactions using a group's color would not update when said group's color was changed. This PR also includes a change that updates the `Ungrouped` row's color when it is changed in the configuration.

Closes #429

### DIFF
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -797,9 +797,8 @@ public class Account : IDisposable
     /// Deletes a group from the account
     /// </summary>
     /// <param name="id">The id of the group to delete</param>
-    /// <param name="defaultTransactionColor">A new transaction color for transactions belonging to deleted group</param>
     /// <returns>(Result, BelongingTransactions)</returns>
-    public async Task<(bool Result, List<uint> BelongingTransactions)> DeleteGroupAsync(uint id, string defaultTransactionColor)
+    public async Task<(bool Result, List<uint> BelongingTransactions)> DeleteGroupAsync(uint id)
     {
         using var cmdDeleteGroup = _database!.CreateCommand();
         cmdDeleteGroup.CommandText = "DELETE FROM groups WHERE id = $id";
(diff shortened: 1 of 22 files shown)


DaPigGuy:
Transaction color should also update if the group gets deleted

nlogozzo:
> Transaction color should also update if the group gets deleted

Done :)

fsobolev (changes requested):
Something goes wrong, can't find where exactly... On group color change the color of belonging transaction gets updated, but also its currency gets changed to unknown

Снимок экрана от 2023-04-09 19-20-47

NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -815,13 +817,19 @@ public async Task<bool> DeleteGroupAsync(uint id)
                 if (pair.Value.GroupId == id)
                 {
                     pair.Value.GroupId = -1;
+                    if (pair.Value.UseGroupColor)
+                    {
+                        pair.Value.UseGroupColor = false;
+                        pair.Value.RGBA = defaultTransactionColor;
```
fsobolev, line 823: do we really need to change RGBA to default here? At least in GNOME RGBA is always saved from ColorDialogButton, even when UseGroupColor is true. It also can be other than the default value if a user changed it, so I think it does make sense to use it without changing to default.

fsobolev (changes requested):

nlogozzo:
> Something goes wrong, can't find where exactly... On group color change the color of belonging transaction gets updated, but also its currency gets changed to unknown

Fixed in the latest commit. Somehow `CultureForNumberString` had a `.Parent` attached to it making it use the parent culture instead of the actual culture itself 😅 VS IntelliSense probably suck it in on me haha

NickvisionMoney.Shared/Models/Account.cs
nlogozzo: Yeah you are right, i'll remove that line.

fsobolev (approved):

PR #434, opened 2023-04-09
Revert "Default to group color on selecting group"

fsobolev:
I disagree with the change in #431. What is the benefit? I don't understand for what scenarios it is an improvement. But what if a user selects unique color first and then selects a group? Or simply wants to change a group for an existing transaction without changing color? Now it requires additional actions to achieve the goal.

### DIFF
--- a/NickvisionMoney.GNOME/Views/TransactionDialog.cs
+++ b/NickvisionMoney.GNOME/Views/TransactionDialog.cs
@@ -231,10 +231,6 @@ public partial class TransactionDialog : Adw.Window
                 {
                     _colorDropDown.SetSelected(1);
                 }
-                else if (!_colorDropDown.GetVisible())
-                {
-                    _colorDropDown.SetSelected(0);
-                }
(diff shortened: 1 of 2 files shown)


nlogozzo:
The way it's implemented...if a user selects a group and then a unique color...and then edits this existing transaction, if the group is changed, the unique color stays.
Same for new transactions...in testing i believe i remember if a new transaction was created and then a unique color ans then a group, the color stays. The if statement in code check if the color button is shown or not, and if it is, it doesn't change anything.

fsobolev:
Okay, sorry, I needed to test it before trying to understand the change and write examples when this is undesired IMO. Still, it is undesired, because my examples are valid if the group that was initially selected is "Ungrouped".

nlogozzo:
> But what if a user selects unique color first and then selects a group?

On further testing your right...this would be very unexpected to the user.

Best to revert.

nlogozzo (approved):

PR #437, opened 2023-04-10
Update Description

nlogozzo:
To be merged after V2023.4.0 stable release.


fsobolev:
Why after and why change long description at all? It matches short description here

nlogozzo:
> Why after

Requires translation change. RC1 signifies string freeze

> why change long description at all? It matches short description here

Ones not really longer then the other, they're the same size. And the "short" one sounds a lot better IMO.

fsobolev:
> Requires translation change. RC1 signifies string freeze

That string existed before RC1, because it was the short description, so translations can be copy-pasted https://github.com/nlogozzo/NickvisionMoney/blob/main/NickvisionMoney.Shared/Resources/Strings.resx#L1146

> And the "short" one sounds a lot better IMO.

Well I agree it sounds better, and it's great short description. But I don't like that we have the same long and short descriptions 😞

nlogozzo:
> That string existed before RC1, because it was the short description, so translations can be copy-pasted

Ah ok, i'll go through translations and update them when I get a chance...won't be till later tomorrow as I have an exam in the morning 🙄

> EDIT: Suggestion for long description (metainfo-only, in doap long description is not necessary I think):

Looks good to me. Feel free to commit it in the metainfo. Feel free to make one for Tube Converter and I'll go through those translations too tomorrow.

fsobolev:
> Looks good to me. Feel free to commit it in the metainfo.

Great, thanks! I will also update `_translate-meta` recipe in justfile, because currently it inserts text using string numbers, which is not the best way to handle it.

nlogozzo:
Ok i updated the descriptions for translations that had `Summary.GTK` translated. If `Summary.GTK` was not translated for a language, that language's translation was left untouched.

nlogozzo:
> EDIT: Suggestion for long description (metainfo-only, in doap long description is not necessary I think):

Committed.

nlogozzo:
Reverted the commit because the script that translates this description needs to be updated to support a new list.

fsobolev:
@nlogozzo metainfo should be good now. I renamed `Summary.GTK` to `MetadataDescription.GTK` because it's the only place when the long description is used (feel free to change if you want to use it for WinUI somewhere), keeping old values except for English and Russian. Script was changed to be more intelligent (it doesn't rely on string numbers or the content of an element, can be a simple text like in most translations now or another tree) and the generated file passes relaxed validation (regular validation gives the same errors as without the changes).

nlogozzo:
Thank you :) Will merge after V2023.4.0 stable is released at 12:00 AM EST tonight since we changed strings. 

Feel free to update the scripts in Application and Tube Converter and add a long description for them

nlogozzo:
https://github.com/NickvisionApps/Denaro/releases/tag/2023.4.1-beta1

PR #445, opened 2023-04-13
WinUI Design Improvements

### LINKED ISSUE #443: WinUI Design Improvements
- Fix sorting submenus (maybe use custom flyout)
- Months in range selectors are not localized
- TransactionDialog not handling custom decimal and group separators
    - > I have custom currency with . decimal and , group separators. A value 1,000.103 is showing fine in transactions list, but when I open the transaction it says that the value is incorrect - group separator is treated as decimal separator, both 1000.103 and 1003,103 are valid and give the same result
- Hovering the ID button of a TransactionRow looses the transaction color
- Ensure luma coloring works for ID button

nlogozzo:
Closes #443


nlogozzo:
Luma For TransactionRow:

image

fsobolev (approved):

PR #448, opened 2023-04-15
Better Export System and Non-Latin Numerals

### LINKED ISSUE #245: Better export
Currently it's only possible to export all transactions sorted by ID. The idea is to have 2 export options: "Export All" and "Export Current". The latter will export only transactions that are currently shown in the window, also sorting should be kept. "Export Current" will also have "New Account" as an option for output format - this way a user will be able to split an account (can be useful in case a user imported a file with a lot of transactions). "Export All" should not have this output option because it's useless.

### LINKED ISSUE #435: Consider using hardcoded or different icon for receipt upload
`document-send-symbolic` doesn't look appropriate with Yaru theme

Снимок экрана от 2023-04-10 02-06-47

### LINKED ISSUE #450: Use non-latin numerals for cultures that have alternative numerals
https://stackoverflow.com/questions/2298153/how-can-i-convert-english-digits-to-arabic-digits - examples how this can be implemented

**EDIT**: [note](https://www.github.com/dotnet/runtime/issues/47749#issuecomment-771815076) about non-ASCII digits formatting

nlogozzo:
Closes #245 
Closes #450
Closes #435 

## TODO
- [x] Backend Implementation
- [x] WinUI Implementation
- [x] GNOME Implementation


nlogozzo:
@fsobolev Could you also quickly implement #435 in this PR?

I'd like to release beta1 once this is merged.

NickvisionMoney.Shared/Helpers/CurrencyHelpers.cs
```diff
@@ -39,6 +39,19 @@ public static string ToAmountString(this decimal amount, CultureInfo culture, bo
...
+                           .Replace("1", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[1])
+                           .Replace("2", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[2])
+                           .Replace("3", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[3])
+                           .Replace("4", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[4])
+                           .Replace("5", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[5])
+                           .Replace("6", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[6])
+                           .Replace("7", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[7])
+                           .Replace("8", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[8])
+                           .Replace("9", CultureInfo.CurrentCulture.NumberFormat.NativeDigits[9]);
+        }
```
fsobolev: This will cause validation of amount values in dialogs to fail, because NativeDigits are not digits for dotnet (see the link I attached in the issue).
Also I think it would be good to have a setting to disable native digits.
I can work on this a bit later.

fsobolev (changes requested):

NickvisionMoney.Shared/Helpers/CurrencyHelpers.cs
nlogozzo: I made it so that input dialogs will use Latin digits and only labels will show native digits. Because it'd be impossible to input and parse non-ASCII native digits according to that comment. 

https://imgur.com/a/hfUGpSe
*I don't have the font to display eastern Arabic digits but it's there*

> Also I think it would be good to have a setting to disable native digits.

Maybe another field in custom currency: `Display Native Digits`

nlogozzo: Requested a review to trigger CI :)

fsobolev: > Maybe another field in custom currency: Display Native Digits

I think it should be global, not per-account. If a user don't want to use native digits they probably don't want to use it everywhere.

> Because it'd be impossible to input and parse non-ASCII native digits according to that comment.

We can create a helper function to convert native digits in string back to latin and then parse to double.

nlogozzo: Global works :)

fsobolev (approved):

nlogozzo:
Going to make it where exporting the current view will also display in the sorted order then I'll merge

fsobolev (approved):

PR #470, opened 2023-04-22
GNOME - Fix crash with unset locale

### LINKED ISSUE #465: Crashes on Opening a File (Aarch64)
When selecting an nmoney file to open, the application crashes. If a new one is created, the application would still crash. This is only happening on my Aarch64 device. On x86_64 everything runs as it should. I tried switching to the beta branch but the error was still there.
Denaro is installed from Flathub and the device is running PostmarketOS.

fsobolev:
Closes #465 
When locale is unknown, it's now set to en_US.
Tested with LC_ALL=C and also in Alpine distrobox.

### DIFF
--- a/NickvisionMoney.GNOME/Program.cs
+++ b/NickvisionMoney.GNOME/Program.cs
@@ -45,7 +45,7 @@ public partial class Program
     /// </summary>
     public Program()
     {
-        if (string.IsNullOrEmpty(CultureInfo.CurrentCulture.ToString()))
+        if (CultureInfo.CurrentCulture.Equals(CultureInfo.InvariantCulture))
         {
             CultureInfo.CurrentCulture = new CultureInfo("en-US"); // Fix #465
         }
--- a/NickvisionMoney.GNOME/Program.cs
+++ b/NickvisionMoney.GNOME/Program.cs
@@ -45,7 +45,11 @@ public partial class Program
     /// </summary>
     public Program()
     {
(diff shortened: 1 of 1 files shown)


badcel:
Wouldn't it be more explicit to check for the [invariant culture](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-7.0) then to fall back to the string level?

fsobolev:
> Wouldn't it be more explicit to check for the [invariant culture](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-7.0) then to fall back to the string level?

I didn't get how to compare current culture with invariant 😅 `CurrentCulture == InvariantCulture` doesn't work, and I didn't manage to google for a proper solution

badcel:
Yeah it's a pretty old API.

`CulureInfo` provides an `Equals` method which checks for the name and the compare info to determine if two cultures are the same. So you would write it like this:

```CS
if(CultureInfo.CurrentCulture.Equals(CultureInfo.InvariantCulture))
{
    ...
}
```

See [dotnet 7 source for culture info](https://github.com/dotnet/runtime/blob/d099f075e45d2aa6007a22b71b45a08758559f80/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L651).

fsobolev:
Thanks @badcel :)

nlogozzo (approved):

PR #482, opened 2023-04-30
Added the snap build files

soumyaDghosh:
Added the snap build files same as the TC


fsobolev:
In TC there's only snap yml without github action. Why so? If action is required to build and upload package, [path](https://github.com/snapcore/action-build#path) can be specified.

soumyaDghosh:
> In TC there's only snap yml without github action. Why so? If action is required to build and upload package, [path](https://github.com/snapcore/action-build#path) can be specified.

Nah, it's just to test if the snap is building or not. I'd add the workflow for TC also.

nlogozzo (approved):

nlogozzo:
> Nah, it's just to test if the snap is building or not. I'd add the workflow for TC also.

Please do :)

Issue #67, opened 2022-11-20
Online Sync


junocomp:
Use ofxtools or something similar to sync to various bank accounts.

nlogozzo:
Money cannot connect to online bank accounts directly as of now, however there has been support for importing data from an ofx file locally for a while.

aronkvh:
that'd be a huge reason to use Denaro for me too

bennypowers:
What about encrypted cloud sync across devices? *not* syncing to the bank's network, but syncing between my manually-entered data across different devices

nlogozzo:
> What about encrypted cloud sync across devices? _not_ syncing to the bank's network, but syncing between my manually-entered data across different devices

This is already possible in Denaro. You can set a password for an account with will encrypt the account file. You can then place that file in Dropbox, OneDrive, MEGA, whatever you use and sync it across devices. All you need is your Denaro account file's password and you can access the file.

bennypowers:
@nlogozzo please clarify.

How does the comment from May 5 address this use case:

I have a GNOME desktop with denaro installed
I have an android phone
My wife has an android phone

I want to collect receipts and enter them manually on my GNOME desktop
I want to add or update entries in our Denaro accounts from my android phone
I want my wife to be able to update entries in our Denaro accounts from her android phone

nlogozzo:
> @nlogozzo please clarify.
> 
> How does the comment from May 5 address this use case:
> 
> I have a GNOME desktop with denaro installed
> I have an android phone
> My wife has an android phone
> 
> I want to collect receipts and enter them manually on my GNOME desktop
> I want to add or update entries in our Denaro accounts from my android phone
> I want my wife to be able to update entries in our Denaro accounts from her android phone

Ah well this is a different case. My comment was treating "different devices" as other PCs, which is where you can just use a cloud storage to sync the .nmoney file. For android devices this isn't possible until we finish #277 . Which we have plans to complete in the coming months...

antrmn:
> > What about encrypted cloud sync across devices? _not_ syncing to the bank's network, but syncing between my manually-entered data across different devices
> 
> This is already possible in Denaro. You can set a password for an account with will encrypt the account file. You can then place that file in Dropbox, OneDrive, MEGA, whatever you use and sync it across devices. All you need is your Denaro account file's password and you can access the file.

That sounds like a weak solution without proper conflict handling

nlogozzo:
[...]

PR #493, opened 2023-05-05
docs: Update README.md

bennypowers:
see https://github.com/NickvisionApps/Denaro/issues/67#issuecomment-1535746116


fsobolev:
This is misleading. There's no encrypted cloud sync in the Denaro. Accounts can be encrypted and so it should be safe to upload them on a cloud, but it's done manually by user, there is no syncing interface in the app.

PR #505, opened 2023-05-23
refactoring account class

rhilliges:
Introduced new AccountRepository class and started moving code. Tested it on my personal nmoney file.

I already finished the refactoring. I thought it might be nicer to have several smaller pull requests, but if you prefer one bigger merge, let me know.

### DIFF
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -168,6 +168,8 @@ public class Account : IDisposable
 
         Metadata = _accountRepository.GetMetadata();
 
+        Groups = _accountRepository.GetGroups();
+
         Transactions = _accountRepository.GetTransactions();
 
         foreach (var transaction in Transactions.Values)
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -7,7 +7,6 @@ using QuestPDF.Drawing;
 using QuestPDF.Fluent;
 using QuestPDF.Helpers;
 using QuestPDF.Infrastructure;
-using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.Formats.Jpeg;
 using System;
 using System.Collections.Generic;
@@ -31,10 +30,9 @@ public enum ExportMode
 /// </summary>
 public class Account : IDisposable
 {
-    private bool _loggedIn;
     private bool _disposed;
-    private SqliteConnection? _database;
-    private bool? _isEncrypted;
+    private SqliteConnection _database { get => _accountRepository.Database; }
+    private AccountRepository _accountRepository;
 
     /// <summary>
     /// The path of the account
@@ -43,27 +41,27 @@ public class Account : IDisposable
     /// <summary>
     /// The metadata of the account
     /// </summary>
-    public AccountMetadata Metadata { get; init; }
+    public AccountMetadata Metadata { get; set; }
     /// <summary>
     /// A map of groups in the account
     /// </summary>
-    public Dictionary<uint, Group> Groups { get; init; }
+    public Dictionary<uint, Group> Groups { get; set; }
     /// <summary>
     /// A map of transactions in the account
     /// </summary>
-    public Dictionary<uint, Transaction> Transactions { get; init; }
+    public Dictionary<uint, Transaction> Transactions { get; set; }
     /// <summary>
     /// Whether or not an account needs to be setup
     /// </summary>
-    public bool NeedsAccountSetup { get; private set; }
-    /// <summary>
+    public bool NeedsAccountSetup { get => _accountRepository.NeedsAccountSetup; }
+      /// <summary>
     /// The next available group id
     /// </summary>
-    public uint NextAvailableGroupId { get; private set; }
+    public uint NextAvailableGroupId { get => _accountRepository.NextAvailableGroupId; }
     /// <summary>
     /// The next available transaction id
     /// </summary>
-    public uint NextAvailableTransactionId { get; private set; }
+    public uint NextAvailableTransactionId { get => _accountRepository.NextAvailableTransactionId; }
     /// <summary>
     /// The income amount of the account for today
     /// </summary>
@@ -84,18 +82,12 @@ public class Account : IDisposable
     /// <param name="path">The path of the account</param>
     public Account(string path)
     {
-        _loggedIn = false;
         _disposed = false;
-        _isEncrypted = null;
         Path = path;
-        Metadata = new AccountMetadata(System.IO.Path.GetFileNameWithoutExtension(Path), AccountType.Checking);
         Groups = new Dictionary<uint, Group>();
-        Transactions = new Dictionary<uint, Transaction>();
-        NeedsAccountSetup = true;
-        NextAvailableGroupId = 1;
-        NextAvailableTransactionId = 1;
         TodayIncome = 0;
         TodayExpense = 0;
+        _accountRepository = new AccountRepository(path);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


nlogozzo:
Thanks for the PR!
I hate to throw away all this work, but on second thought I don't think I'll accept this PR. While I do agree the Account class is huge, I don't think separating it into `AccountRepository` solves the issue as it create more de-fragmentation. The code should stay together since it all has to do with an Account model.

rhilliges:
Thanks for your response @nlogozzo, maybe I'll explain a little more, why I think that this is actually a good idea. :)
Your point is that the code should stay together, because it all has to do with the account model. The way models should work is, that they should only contain business/domain logic and should be independent of the underlying infrastructure as much as possible. The infrastructure in case of Denaro is the SQLite database. The AccountRepository separates the business logic concern from the infrastructure concern and gives flexibility, if at any point in time Denaro wants to support another storage option (json, xml, MySQL ...). That may never happen, but having this flexibility is still gonna pay off in the long run, trust me ;)

nlogozzo:
I can see how separating into `AccountRepository` would be useful if we wanted to switch storage options...but I doubt that would ever happen as it would require a lot of code to support migrating to a new storage model if we so choose. 

However, AddTransaction, UpdateTransaction, DeleteTransaction, all those methods' logic should be moved to `AccountRepository` too, no? Since they are all sqlite code. 
And at that point...we just moved all `Account` code to `AccountRepository` so we should just use that instead of `Account`, which brings us back to one model where we are now.

nlogozzo:
I feel as switching to another storage option, which we have no plans of doing anytime soon, will cause major rewrites whether it be in `Account` itself or `AccountRepository` if we move everything there.

rhilliges:
The final result would look something like this:
https://github.com/rhilliges/Denaro/tree/feature/refactoring-account-class-v5
I just mentioned the storage as an example. Another good argument would be code testability or finding bugs.

rhilliges:
The account class would still have all those methods, but would delegate storage logic to the repository.

nlogozzo:
The problem is that while creating these separate `CsvExporter` and `AccountRepository` classes do indeed make it a little easier to read, all of this functionality is only used by the `Account` model and thus should live in that model and can thus be tested in that model. 

It's different than something like `AccountMetadata` that lives separately from `Account`, because `AccountMetadata` is used  by Recent Accounts and Transfers without loading the whole `Account` model, thus it makes sense to separate that logic because it's used by other components then just `Account` itself.

nlogozzo:
So creating separate `CsvExporter`  and `AccountRepository` classes and other fragmentations of `Account` make it harder to maintain `Account` model since it is now separated into all these different pieces and places with no benefit since they are all only used by `Account` to begin with. 

@fsobolev Thoughts? Maybe you can explain what I mean better....

rhilliges:
I would like to focus this discussion on the `AccountRepository`. I just shared the code to address your concern that all code would be moved over to the `AccountRepository` class. The `Account` still has over 1000 lines of code even after separating the database.

You actually made a good point, that the split makes it easier to read. In my opinion readability goes hand in hand with maintainability. As the code base grows its gonna be increasingly harder to understand all the code, especially if everything lives in a single file.

nlogozzo:
> In my opinion readability goes hand in hand with maintainability

Yes but we shouldn't make things more readable to "increase maintainability" it doesn't happen like that. Separating all this functionality into other classes when it's all related to one model and used be one model, will cause fragmentation and hurt maintainability.

> As the code base grows its gonna be increasingly harder to understand all the code, especially if everything lives in a single file.

I disagree. What makes it more maintainable and easier to follow is separation into methods, keeping everything in one file isn't an issue, as long as functionality is organized into it's own methods and separated as it is now in `Account`.

rhilliges:
Ok, so if all methods from `AccountRepository` would live in the `Account` as private methods, you would be ok with that?

nlogozzo:
They already do, that's what I'm trying to say.

`AccountRepository.Login()` is already in `Account.Login()` no need to move it to `AccountRepository` for `Account` to just call that method. Just keep it in `Account`.

rhilliges:
Ok with `Login` you are right, but what about `GetMetadata` for example or `GetGroups`?

nlogozzo:
Something like the TranascationFilter class up made I'd accept because that is abstracting a bunch of random variables everywhere into a nice filter API.

fsobolev:
> @fsobolev Thoughts? Maybe you can explain what I mean better....

I think you explained well: separating things used by only one model doesn't follow your position on what is good for maintainability.
Although there's one more point:

> The AccountRepository separates the business logic concern from the infrastructure concern and gives flexibility, if at any point in time Denaro wants to support another storage option (json, xml, MySQL ...).

If there would be a need to bring some breaking changes so that Denaro would need to work with 2 different formats or 2 different database layouts, that would make sense. But this is unlikely to happen... Probably will never happen. Current format works well, there's no point in removing features (even if something wouldn't be implemented in UI, it doesn't need to be removed from backend) and all issues that we had were solved in the past. And since we are talking about format of the main data files, breaking changes to it should be avoided at all to not cause compatibility issues in case `nmoney` files will become supported by other apps.
So introducing changes to code to get flexibility that will never be used is fixing what works well.
I'm sorry that I give arguments to throw away the work, but I can't find benefits unfortunately.

rhilliges:
> They'll be the same case as login. There is no need to move/fragment it to AccountRepository for Account to just call that method. Just keep it in Account.

I meant what if those methods would be private methods inside of `Account`? Would you accept a pull request that separates the database logic inside the same class? It would still increase readability, but wouldn't add fragmentation

nlogozzo:
If you make a private method in the Account class with the database logic, and then in the public method call that private method,  that's fragmented. Just keep it all in the public method what's the point to separate it.

rhilliges:
It would still increase readability quite a bit and makes the code easier to understand, which is nice for potential new contributors.
And in my opinion it still makes the application easier to maintain. If there is a bug in the domain logic, you don't have to dig through all the database code and vice versa.

nlogozzo:
> It would still increase readability quite a bit and makes the code easier to understand, which is nice for potential new contributors.
> And in my opinion it still makes the application easier to maintain. If there is a bug in the domain logic, you don't have to dig through all the database code and vice versa.

I disagree because like this you'll have to search through whatever the problem is a database problem or domain logic problem, whereas if it's all one method you know the problem lives there. Furthermore, it doesn't add that much readability improvmenet because it all lives under it's method. AddTranascationAsync contains all the code it needs. IMO it would make it harder to understand if there's now two methods AddTransacationAsync and a private AddTransactionDatabaseAsync where people now need to learn the difference between the two methods whereas if it's all as one they know the code all works together.

rhilliges:
I think the ``LoadAsync`` method would be the example I would pick. Refactored it would look something like this:
```csharp
public async Task<bool> LoadAsync()
    {
        if (!LoggedIn)
        {
            return false;
        }
        MigrateDatabase();

        Metadata = GetMetadata();

        Transactions = GetTransactions();

        foreach (var transaction in Transactions.Values)
        {
            if (transaction.Date <= DateOnly.FromDateTime(DateTime.Now))
            {
                var groupId = transaction.GroupId == -1 ? 0u : (uint)transaction.GroupId;
                Groups[groupId].Balance += (transaction.Type == TransactionType.Income ? 1 : -1) * transaction.Amount;
                if (transaction.Type == TransactionType.Income)
                {
                    TodayIncome += transaction.Amount;
                }
                else
                {
                    TodayExpense += transaction.Amount;
                }
            }
        }
        await SyncRepeatTransactionsAsync();
        FreeMemory();
        return true;
    }
```
IMO this is a lot easier to understand and would follow the ``Clean Code`` principles (its a good read if you wanna check it out). 
Just to name a few examples, where this structure would be helpful besides readability:
If there is a problem migrating the database, i know exactly where to look. If something is off with the metadata, the method name is right there. Same goes for if a transaction does not show up in the view, but I can find it in the database.

nlogozzo:
Yes but the MigrateDatabase() and GetMetadata() and GetTransactions() functions will only be called once from the `LoadAsync` method, as that's the only place they need to be used, so no point in fragementing them into other functions, just keep the code in the `LoadAsync` function since that's the only place it's used. 

Functions shouldn't be made just to make code more readable, they are used to reduce code duplicating and structure the API. `LoadAsync` is used by controller so it needs to be a function in `Account`, but there is no need to fragment the code inside it into other functions if they are only used by `LoadAsync`

rhilliges:
``Functions shouldn't be made just to make code more readable``

I think this is the main thing where I disagree. As someone who has several years of experience I would do it differently, but in the end its still your app!

nlogozzo:
Yeah i'd like to keep it all together. But I saw another branch where you made that `TranascationFilter` class...that i'd be open to accepting

PR #506, opened 2023-05-23
Remove WinUI + Switch To Cake

nlogozzo:


nlogozzo:
@fsobolev Could you update the cake script to install the [mimeinfo](https://github.com/NickvisionApps/Denaro/blob/no-windows/NickvisionMoney.GNOME/org.nickvision.money.extension.xml.in) and the docs? Then i'll merge.

PR #517, opened 2023-05-25
groups totals based on shown transactions

### LINKED ISSUE #366: Show Amounts Based On Transactions Shown
From @fsobolev :
> What would be nice is if overview (and maybe groups too) would show numbers for filtered transactions. I think this can be considered, it doesn't add any complexity for users to work with, but we need to decide how the app should treat future transactions in this case. @nlogozzo

So basically I think it will show the number of transactions shown in the transactions list (including future transactions if they are shown in the list)

We can add a row underneath the `Expense` row in the overview section to show the `# of Transactions Shown: _` and then we can find a place for each group row to show the number of transactions shown in the list that belong to that group.

rhilliges:
Closes #366 

- introduced new TransactionFilter
- pass TransactionFilter to Account
- pass TransactionFilter from Account to Group
- calculate Group Balance inside of Group

### DIFF
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -1084,6 +1084,7 @@ public class AccountViewController : IDisposable
 
     private void GroupRowUIUpdate(uint groupId)
     {
+        _account.Groups[groupId].UpdateBalance();
         GroupRows[groupId].UpdateRow(_account.Groups[groupId], GroupDefaultColor, CultureForNumberString, _transactionFilter.IsFilterActive((int)groupId));
     }
 }
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -18,7 +18,7 @@ public class AccountViewController : IDisposable
     private bool _isOpened;
     private bool _disposed;
     private readonly Account _account;
-    private readonly TransactionFilter transactionFilter;
+    private readonly TransactionFilter _transactionFilter;
 
     /// <summary>
     /// Gets the AppInfo object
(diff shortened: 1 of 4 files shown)


rhilliges:
@nlogozzo 
The general implementation is done. For some reason the UI does not get updated yet. Maybe you see a reason why?
Feel free to give me feedback, the code probably won't change much anymore.

nlogozzo:
> For some reason the UI does not get updated yet. Maybe you see a reason why?

I'll take a look at it in a little bit

In the mean time, could you target this PR towards the [V2023.6.0](https://github.com/NickvisionApps/Denaro/tree/V2023.6.0) branch since `main` is still on development of `V2023.5.0`?

nlogozzo:
I don't like the fact that `Transaction`s are being stored in the group themselves. Recalculating the `Balance` property of the group every-time it is accessed is something that should be avoided for performance concerns.

rhilliges:
I'm still debugging at the moment. The balance is not being updated at all.

rhilliges:
> The entire balance of each group is still recalculated every-time `UpdateBalance` is called. In the current stable implementation, the entire balance is calculated when the account is opened, and then it is modified individual as transactions are added and removed from the groups, thus avoiding an entire recalculation.

I changed the code so on startup the group balance is modified with each new transaction.

> Take an Account with 2000+ transactions. For every group it'll have to iterate over those 2000+ transactions and recalculate the group balance everytime, this will get pretty slow really fast. And it's something we need to keep in mind as we do have users who have real accounts with 1000+ transactions.

With the current implementation, a group only has references to the transactions that actually belong to it. Since a transaction has only one group, every transaction would only be iterated over once.


To keep performance in mind, I would suggest on putting a delay on the search field (300ms or something like that), so the recalculation/filtering does not get triggered with each keystroke.

rhilliges:
@nlogozzo 
PR is ready :)

nlogozzo:
I merged the V2023.6.0 branch into main. If you could make a new PR basing this feature against `main` branch now, then I'll review it

PR #520, opened 2023-05-26
Cake - Set RID when publishing

### LINKED ISSUE #518: Cake builds for every architecture
Hi, I'm working on updating the aur package. I used the same commands for denaro and tube-converter but while tube-converter behaves correctly, in denaro the build system installs runtimes for every architecture. I'm using these commands:

```
dotnet tool restore
dotnet cake --target=Publish --prefix=/usr --ui=gnome
dotnet cake --target=Install --destdir="$pkgdir"
```

fsobolev:
Closes #518

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -47,18 +47,26 @@ Task("Publish")
     .Does(() =>
 {
     var selfContained = Argument("self-contained", false) || HasArgument("self-contained") || HasArgument("sc");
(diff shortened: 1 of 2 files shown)


nlogozzo:
Could you make these changes to `Tube Converter` and `Application` just so we can ensure this doesn't happen again?

PR #526, opened 2023-05-27
groups totals based on shown transactions

rhilliges:
continues #517 

@nlogozzo 
ready for review :)

### DIFF
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -1084,6 +1084,7 @@ public class AccountViewController : IDisposable
 
     private void GroupRowUIUpdate(uint groupId)
     {
+        _account.Groups[groupId].UpdateBalance();
         GroupRows[groupId].UpdateRow(_account.Groups[groupId], GroupDefaultColor, CultureForNumberString, _transactionFilter.IsFilterActive((int)groupId));
     }
 }
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -18,7 +18,7 @@ public class AccountViewController : IDisposable
     private bool _isOpened;
     private bool _disposed;
     private readonly Account _account;
-    private readonly TransactionFilter transactionFilter;
+    private readonly TransactionFilter _transactionFilter;
 
     /// <summary>
     /// Gets the AppInfo object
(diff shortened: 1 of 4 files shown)


rhilliges:
@nlogozzo 
still interested in the PR? Gonna start my new job soon and probably won't have much time for fixes then

nlogozzo:
Yes. I've been focused on getting Tube Converter V2023.6.0 released. Now that it has been released I'l be shifting my work back towards Denaro. I'm working tomorrow and Saturday, so probably won't get to this until Sunday :)

rhilliges:
Alright, sounds good :)

nlogozzo:
Ok, finally got time to test this PR...and it works, but it doesn't do what #366 asked. The issue was asking to change the amounts shown in the UI sidebar to reflect the amounts of the transactions actually shown in the list.

You basically just rewrote the way in which the group amounts were calculated, which is fine, but your implement made the calculation slower then it is currently implemented in `main`.
Take this example: 
- A group has 5 transactions 
    - A total balance of $25
- A new transaction of $7 is added to the group 

Now let's look at how the current implementation in `main` preforms vs. your changes in this PR:
- Currently, in `main`, the `Account` model manages each group's balance and updates it accordingly when transactions are added or removed from the group. So in the above example, when the new $7 transaction is added, the model will:
    - Do $25+$7 and set the new balance to $32
        - a basically instant calculation 
    - And we're done.

- With your changes, it's a lot slower. You have the `Group` model store all the transactions and then when one is added or removed, that list is updated and the entire list is iterated to recalculate the balance. So let's use the same example as above. When the new $7 transaction is added, your model will:
    - Add that transaction to the group's internal list
       - Shouldn't take up too much time to do, but we have higher RAM usage as each group now has a list of transactions
   - Iterate over the entire list (using `Sum`, which is basically a `for` loop) and recalculate the balance
       - This is wayyy to slow because now we have to do, for example, $32+$5-$7-$2-$3 = $25 and finally +$7 = $32
       - Now this is only with 5 existing transactions, imagine if we had 1000 in a group. We'd have to iterate over 1000 transactions and re-sum the entire balance for no reason and with 1000 transactions, this can definitely take a few seconds to complete
     - Compared to our implementation in `main`, no matter how many existing transactions there are, 1000 or 10000, it will just do one single calculation: Current_Group_Balance + New_Transaction_Amount . No recalculation or reiteration required, saving a lot of time. Plus the Group model doesn't have a list of Transactions to store and manage, this saving RAM as well.
     
Therefore, because of the performance issues mentioned above and the fact that this doesn't implement #366, I will be closing this PR. However, I thank you for your time in creating this PR and learning the Denaro codebase. 

*As I've said before, I do like the idea of the `TransactionFilter` model you have and would love to have that in the codebase if you'd like to make a new PR for it.*

PR #531, opened 2023-05-28
GNOME Circle Tweaks

### LINKED ISSUE #452: GNOME Circle Tweaks
https://gitlab.gnome.org/Teams/Circle/-/issues/143#note_1723472

- [x] Implement Enter and ESC keys in dialogs
- [x] Remove accounts from Recents list
- [x] Arbitrary color labels don't work
    - image
    -  image
- [x] Those Rows are too busy. Maybe remove icons, move delete into edit dialog. 
    - image
- [x] Don't use color to mark today in calendar (at least not yellow)
- [x] Use AccountSetupDialog for selecting filename and save location of account
- [x] Use correct minus sign (#525)
- [x] Switch sorting default to by date, recent first
- [x] "tbh, I would drop this "color per transaction" concept completely. Colors as indicators don't work for everyone. And even if your eyes can differentiate colors, you have to remember what they mean. Just replace it with labels which have colors. And then check if you really want to go with groups + labels. But color + labels + groups is just a bit much."

nlogozzo:
Closes #452


nlogozzo:
@fsobolev If you could take a look at the PR and see what you think so far. I finished most things from the circle review. I'm going to work on "Use AccountSetupDialog for selecting filename and save location of account" next. 

Would you be able to do "Arbitrary color labels don't work" and "Don't use color to mark today in calendar (at least not yellow)" for me?

nlogozzo:
The only thing I don't like is having three buttons in the `TransactionDialog`:
image
it becomes too wide. Maybe we can move `Make a Copy` as an activatable row similar to `Extras`? Or just move the button somewhere else.

fsobolev:
> Or just move the button somewhere else.

To headerbar maybe?

nlogozzo:
Maybe in like the top left where a back button would usually be?

And when the back button is needed, when we move to the Extras page, we'll just hide it... or maybe just in the top right next to the close button?

nlogozzo:
Also, when Groups are hidden the `Show` button becomes raised as it is `toggled`. I tried removing the `toggle` CSS class, but no effect. Maybe you can do some CSS?
image

nlogozzo:
> And when the back button is needed, when we move to the Extras page, we'll just hide it.

image
image

fsobolev:
> And when the back button is needed, when we move to the Extras page, we'll just hide it.

Looks good, I like it.

nlogozzo:
PR Ready with NewAccountDialog:
image
image
image
image

@fsobolev Maybe you can make it where when the user is not on the Currency page, the dialog doesn't take with that much vertical space...otherwise it's all ready :)

NickvisionMoney.GNOME/Views/NewAccountDialog.cs
```diff
@@ -0,0 +1,369 @@
...
+    /// <param name="controller">NewAccountDialogController</param>
+    /// <param name="parent">Gtk.Window</param>
+    public NewAccountDialog(NewAccountDialogController controller, Gtk.Window parent) : this(Builder.FromFile("new_account_dialog.ui"), controller, parent)
+    {
+    }
+    
+    /// <summary>
+    /// Presents the dialog
+    /// </summary>
+    public new void Present()
```
DaPigGuy, line 193: Why are we overriding here?

NickvisionMoney.GNOME/Views/MainWindow.cs
```diff
@@ -456,8 +475,6 @@ private bool OnCloseAccountPage(Adw.TabView view, Adw.TabView.ClosePageSignalArg
             _viewStack.SetVisibleChildName("pageNoAccounts");
             _accountMenuButton.SetVisible(false);
             _flapToggleButton.SetVisible(false);
-            UpdateRecentAccountsOnStart();
-            _startPageRecentAccountsGroup.SetVisible(true);
```
fsobolev, line 460: Shouldn't this be kept? The purpose of this wasn't to remove accounts from list, but reorder them if needed.

fsobolev (changes requested):

NickvisionMoney.GNOME/Blueprints/account_settings_dialog.blp
```diff
@@ -57,7 +57,6 @@ Adw.Window _root {
 
                 Adw.ComboRow _accountTypeRow {
                   title: _("Account Type");
-                  subtitle: _("This is only a useful label that doesn't affect how the app works.");
```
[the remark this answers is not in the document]
nlogozzo: Because now that we have the New Account setup dialog, this description is exposed there ans the user can learn that it's just for cosmetic purposes...now real purpose to have it in the Settings dialog itself

NickvisionMoney.GNOME/Views/MainWindow.cs
nlogozzo: I cleaned-up the code when rewriting the recent accounts, so this is no longer needed and it should still reorder like before.

fsobolev (approved):

PR #534, opened 2023-05-29
Fixed metadata appstream not parsed

soumyaDghosh:


snap/snapcraft.yaml
```diff
@@ -1,11 +1,7 @@
 name: denaro # you probably want to 'snapcraft register <name>'
 base: core22 # the base snap is the execution environment for this snap
-version: '2023.5.0' # just for humans, typically '1.2+git' or '1.3.2'
```
nlogozzo, line 3: Why remove the `version`?

soumyaDghosh: It's metadata time :muscle:

nlogozzo:
Why was it closed?

PR #552, opened 2023-06-06
merge main to snap

soumyaDghosh:


fsobolev:
Next time you can simply do

```
git rebase main
git push --force-with-lease
```

So you will not need to open a PR.

soumyaDghosh:
I was searching for those commands, all around the internet!!! Thanks!

Issue #546, opened 2023-06-04
Program crashes when mass importing transactions


NotYourAlejandro:
### Description
I encountered a program crash while attempting to mass import transactions in the application from a csv file. My csv file contains 1030 transactions. The data is sourced from my bank and was manipulated by myself to allow the application to read it successfully.
From my testing, the program's memory usage starts at 153 MiB, then maxes out during the import at 339 MiB.

### Logs
`Process terminated. A callback was made on a garbage collected delegate of type 'NickvisionMoney.GNOME!NickvisionMoney.GNOME.Views.AccountView+GSourceFunc::Invoke'.`

`app-flatpak-org.nickvision.money-60024.scope: Consumed 58.645s CPU time.`

### Steps to Reproduce:
1. Obtain or create a large csv file that is compatible with Denaro.
2. Launch the application.
3. Select an account (preferably an empty one for testing).
4. On the left-hand panel, click `Actions`, then `Import from File`.
5. Choose the large csv file.

### Expected Behavior
The application should import the transactions within the csv file, then displays the results of the import.

### Actual Behavior
The application imports the transactions within the csv file, but crashes when it finishes, failing to display the results. The imported contents exist on next run.

### Software Versions and Hardware
OS: Fedora 38
DE: KDE Plasma 5.27.5
Kernel: 6.2.15
CPU: AMD Ryzen 5 3600X
GPU: NVIDIA GeForce GTX 1080
Memory: 16 GB (7 GB used during testing)

### Attachments
[https://youtu.be/dmZ2ne93Ksc](https://youtu.be/dmZ2ne93Ksc)

nlogozzo:
> Process terminated. A callback was made on a garbage collected delegate of type 'NickvisionMoney.GNOME!NickvisionMoney.GNOME.Views.AccountView+GSourceFunc::Invoke'.

@fsobolev I thought we were done with these errors 😅

fsobolev:
> > Process terminated. A callback was made on a garbage collected delegate of type 'NickvisionMoney.GNOME!NickvisionMoney.GNOME.Views.AccountView+GSourceFunc::Invoke'.
> 
> @fsobolev I thought we were done with these errors sweat_smile

We were done in TC, but Denaro still has weird callbacks :(

fsobolev:
> I can't seem to find them...it looks ok to me. Do you know where they are? I'll fix them

I think the problem is with dynamic callbacks in `CreateGroupRow` etc, where they are recreated each time:

[...]

PR #557, opened 2023-06-07
Show amounts based on transactions shown + Sidebar Changes

### LINKED ISSUE #366: Show Amounts Based On Transactions Shown

### LINKED ISSUE #503: Toggle selection of all groups
From #348:

> Turn the "Reset Groups Filter" into a toggle button that enables/disables all groups. This way I need at most 3 clicks, if I want to see the transactions of one specific group. Right now, I have to disable all other groups one by one.

I don't think the clear groups button should become a toggle, but a new button should be added. Not sure where to place it, but I will think about this.

nlogozzo:
Closes #366 
Closes #503


nlogozzo:
Ok, PR ready!
Code is a little messy, but it will get cleanup up in the rewrite to fix #546 ... either way the feature is there

nlogozzo:
I think maybe, when a group is unchecked, and when income/expense is unchecked, the amount label should be hidden, that way the user doesn't think that the group's transactions were deleted or that the balance is actually `0`, just the transactions are not filtered so there is no amount for it, thus it should be hidden

nlogozzo:
> I think maybe, when a group is unchecked, and when income/expense is unchecked, the amount label should be hidden

Implemented with https://github.com/NickvisionApps/Denaro/pull/557/commits/535d04524de6dcc92bc13ec532f900ac7e4cf97d

image

Let me know what you think or if we should revert it

nlogozzo:
Ok i also updated the sidebar design to add #503 as well as just a better design:
image

Now the PR is ready haha :)

DaPigGuy (approved):

PR #563, opened 2023-06-09
Cake - Install docs and mime

fsobolev:
Turns out docs and mime were missing 😅 

Thanks @soumyaDghosh for pointing out


soumyaDghosh:
LOL!!! Is that always me!

PR #590, opened 2023-06-29
Translation strings fixes

fsobolev:
1. Replaced `&#xA;` with `\n` in one string
2. Added its rules for metadata translation to avoid generating translation for "Nickvision" (default rules file force creates translation for `developer_name`)


nlogozzo:
> Added its rules for metadata translation to avoid generating translation for "Nickvision" (default rules file force creates translation for developer_name)

Do we need this in all other apps?

PR #628, opened 2023-08-09
Tags

### LINKED ISSUE #101: Implement tags for transactions
It would be nice to have the ability to add multiple groups to transactions.

For example a transaction could be in the group "recurring payments" and "donations".
This would really help to organize transactions, which fit more than one group.

fsobolev:
Closes #101 

Tags are strings meant to be used as additional filtering method. A transaction can have as many tags as the user wants. Tags are only stored in transactions, list of tags in an account is generated on loading transactions; as result, there's no need to remove tags - unused ones automatically disappear on the account closing. Tags can contain any characters except `,` because it's used as separator.
`Gtk.ScrolledWindow` with fixed size is used because `propagate-natural-height` doesn't work in a popover, and I didn't find a better way to create tags managing interface without sacrificing usability. It has `card` style class because we need to visually separate it from Add tag entry/button, separator doesn't work here because elements a child of a popover always have margins. Tags are just `Gtk.ToggleButton`s without any additional styling (I tried to make them more rounded first, but I think they looked worse).

Implemented: adding, selecting, loading and saving tags in transactions, as well as exporting/importing tags.

TODO:
- [x] Filtering using tags
- [x] Show/hide tags
- [x] Update po(t)
- [x] Update docs

Unrelated, but I wish to implement in the same PR since I already start polishing some things:
- [x] Saving graphs state (visible/hidden)

### DIFF
--- a/NickvisionMoney.GNOME/Controls/TagButton.cs
+++ b/NickvisionMoney.GNOME/Controls/TagButton.cs
@@ -1,0 +1,28 @@
+using System;
+
+namespace NickvisionMoney.GNOME.Controls;
+
+public class TagButton : Gtk.ToggleButton
+{
+    public int Index;
+    
+    public event EventHandler<int> OnAddTag;
+    public event EventHandler<int> OnRemoveTag;
+    
+    public TagButton(int index, string tag)
+    {
+        Index = index;
+        SetLabel(tag);
+        OnToggled += (sender, e) =>
+        {
+            if (GetActive())
+            {
+                OnAddTag?.Invoke(this, Index);
+            }
+            else
+            {
+                OnRemoveTag?.Invoke(this, Index);
+            }
+        };
+    }
+}
--- a/NickvisionMoney.GNOME/Controls/TagButton.cs
+++ b/NickvisionMoney.GNOME/Controls/TagButton.cs
@@ -2,27 +2,33 @@ using System;
 
 namespace NickvisionMoney.GNOME.Controls;
 
+/// <summary>
+/// Tag toggle button
+/// </summary>
 public class TagButton : Gtk.ToggleButton
 {
-    public int Index;
+    /// <summary>
+    /// Tag index in a list of all tags in an account
+    /// </summary>
+    public int Index { get; init; }
+    /// <summary>
+    /// Tag string
+    /// </summary>
+    public string Tag { get; init; }
     
-    public event EventHandler<int> OnAddTag;
-    public event EventHandler<int> OnRemoveTag;
+    /// <summary>
+    /// Occurs when toggle state has changed
+    /// </summary>
+    public event EventHandler<(int Index, bool Filter)> FilterChanged;
     
+    /// <summary>
+    /// Construct TagButton
+    /// </summary>
     public TagButton(int index, string tag)
     {
         Index = index;
+        Tag = tag;
         SetLabel(tag);
-        OnToggled += (sender, e) =>
-        {
-            if (GetActive())
-            {
-                OnAddTag?.Invoke(this, Index);
-            }
-            else
-            {
-                OnRemoveTag?.Invoke(this, Index);
-            }
-        };
+        OnToggled += (sender, e) => FilterChanged?.Invoke(this, (Index, GetActive()));
     }
 }
--- a/NickvisionMoney.GNOME/Controls/TagButton.cs
+++ b/NickvisionMoney.GNOME/Controls/TagButton.cs
@@ -8,10 +8,6 @@ namespace NickvisionMoney.GNOME.Controls;
 public class TagButton : Gtk.ToggleButton
 {
     /// <summary>
-    /// Tag index in a list of all tags in an account
-    /// </summary>
-    public int Index { get; init; }
-    /// <summary>
     /// Tag string
     /// </summary>
     public string Tag { get; init; }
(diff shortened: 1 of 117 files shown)


NickvisionMoney.GNOME/Controls/TagButton.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+
+namespace NickvisionMoney.GNOME.Controls;
+
+public class TagButton : Gtk.ToggleButton
+{
+    public int Index;
```
nlogozzo, line 7: `{ get; init; }`

NickvisionMoney.Shared/Models/Transaction.cs
```diff
@@ -170,9 +176,10 @@ public object Clone()
             GroupId = GroupId,
             RGBA = RGBA,
             UseGroupColor = UseGroupColor,
-            Receipt = Receipt != null ? Receipt.Clone((x) => { }) : null,
+            Receipt = Receipt?.Clone((x) => { }) ?? null,
             RepeatFrom = RepeatFrom,
             RepeatEndDate = RepeatEndDate,
+            Tags = Tags,
```
nlogozzo, line 182: This needs to be properly cloned. 
```
Tags = Tags.Select(x => x).ToList(),
```

nlogozzo (changes requested):

NickvisionMoney.Shared/Models/Transaction.cs
fsobolev: `Tags = Tags.Select(x => x).ToList(),`

or 

`Tags = new List<string>(Tags),`

? I guess both works, just want to know if there's a preference

NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -793,8 +792,6 @@ public async Task<bool> AddTransactionAsync(Transaction transaction)
         cmdAddTransaction.Parameters.AddWithValue("$useGroupColor", transaction.UseGroupColor);
         cmdAddTransaction.Parameters.AddWithValue("$notes", transaction.Notes);
         cmdAddTransaction.Parameters.AddWithValue("$tags", string.Join(',', transaction.Tags));
-        Tags.AddRange(transaction.Tags.Where(t => !Tags.Contains(t)));
```
nlogozzo: Why was this removed? Shouldn't it be kept if new tags are added to a transaction that don't exist in an account?

NickvisionMoney.Shared/Controllers/AccountViewController.cs
```diff
@@ -481,6 +496,13 @@ public async Task StartupAsync()
                 _filters.Add((int)pair.Value.Id, true);
                 GroupCreated?.Invoke(this, new ModelEventArgs<Group>(pair.Value, null, true));
             }
+            //Tags
+            TagCreated?.Invoke(this, new ModelEventArgs<string>(_("Untagged"), -1, true));
```
nlogozzo, line 500: and make this position `0` and then `+ 1` to all `_account.Tags.IndexOf(tag)`

NickvisionMoney.Shared/Controllers/AccountViewController.cs
```diff
@@ -1059,6 +1108,61 @@ public void UnselectAllGroupsFilter()
...
     }
 
+    /// <summary>
+    /// Updates whether or not a tag filter is active
+    /// </summary>
+    /// <param name="index">Index of the tag</param>
+    /// <param name="active">Whether or not the tag filter is active</param>
+    public void UpdateTagFilter(int index, bool active)
+    {
+        if (index > -1)
```
nlogozzo, line 1118: with above changes make this `if(index != 0)`

NickvisionMoney.Shared/Controllers/AccountViewController.cs
```diff
@@ -1059,6 +1108,61 @@ public void UnselectAllGroupsFilter()
...
+        FilterUIUpdate();
+    }
+
+    /// <summary>
+    /// Reset tags filter, setting all tags to enabled state
+    /// </summary>
+    public void ResetTagsFilter()
+    {
+        _filterUntagged = true;
+        TagUpdated?.Invoke(this, new ModelEventArgs<string>(_("Untagged"), -1, true));
```
nlogozzo, line 1142: `0`

NickvisionMoney.Shared/Controllers/AccountViewController.cs
```diff
@@ -1089,6 +1193,14 @@ private void FilterUIUpdate()
             {
                 continue;
             }
+            if (!_filterUntagged && pair.Value.Tags.Count == 0)
+            {
+                continue;
+            }
+            if (!_filterTags.Intersect(pair.Value.Tags).Any() && pair.Value.Tags.Count > 0)
```
nlogozzo, line 1200: I'd change
```csharp
!_filterTags.Intersect(pair.Value.Tags).Any()
```
to
```csharp
!pair.Value.Tags.Any(x => _filterTags.Contains(x))
```

nlogozzo (changes requested):

NickvisionMoney.Shared/Models/Account.cs
fsobolev: We do this in controller, and by doing this we ensure that a new TagButton gets added in the UI

nlogozzo:
With latest commits, TransactionDialog is broken...working to fix it

nlogozzo:
Ok fixed

nlogozzo:
@fsobolev Implemented showing and hidding tags list and updated po(t)....all that's left is if you can update docs :)

nlogozzo (approved):

PR #654, opened 2023-08-19
ui: Mark 3 strings translatable

sabriunal:
- Enter a tag name
- PDF
- CVS

### DIFF
--- a/NickvisionMoney.GNOME/Blueprints/account_view.blp
+++ b/NickvisionMoney.GNOME/Blueprints/account_view.blp
@@ -11,14 +11,14 @@ menu actionsMenu {
       label: _("Export to File");
       
       submenu {
-        label: "CSV";
+        label: _("CSV");
         
         item(_("Export All Information"), "account.exportToCSVAll")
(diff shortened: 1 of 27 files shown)


nlogozzo:
@fsobolev Could you run GeneratePot and UpdatePo?

sabriunal:
I'm using the flatpak-builder. Which commands do I need to run?

Could you please provide updated instructions for it?

nlogozzo:
You need to have `dotnet` 7 installed...then from the root of the repo run:
- `dotnet tool restore`
- `dotnet tool install --global GetText.NET.Extractor` (which you might need .NET 6 for this to install...which is fine you can install 6 and 7 side by side)
- `dotnet cake --target=GeneratePot && dotnet cake --target=UpdatePo` (this will make the new .pot file and apply it to existing .po files)

nlogozzo (approved):

sabriunal:
It seems a lot of work. Thanks for updating them.

nlogozzo:
> It seems a lot of work. Thanks for updating them.

😅 that's why we just take care of it haha

PR #663, opened 2023-08-23
Amount Display Style Option in Custom Currency

### LINKED ISSUE #644: Option to have a space between the symbol and amount
When using a 3 character currency "symbols" (just letters), Denaro doesn't put a space in-between this symbol and the amount. So, for example, you would see USD1 rather than the correct USD 1. For this reason, Denaro should allow 4 characters as a symbol, or add an option to have a space between the symbol and amount.

nlogozzo:
Closes #644 

image

### DIFF
--- a/NickvisionMoney.Shared/Models/AccountMetadata.cs
+++ b/NickvisionMoney.Shared/Models/AccountMetadata.cs
@@ -41,7 +41,7 @@ public enum RemindersThreshold
 /// </summary>
 public class AccountMetadata : ICloneable
 {
-    private int _customCurrencyAmountStyle;
+    private int? _customCurrencyAmountStyle;
     
     /// <summary>
     /// The name of the account
@@ -112,7 +112,7 @@ public class AccountMetadata : ICloneable
         UseCustomCurrency = false;
         CustomCurrencySymbol = null;
         CustomCurrencyCode = null;
-        CustomCurrencyAmountStyle = 0;
+        CustomCurrencyAmountStyle = null;
         CustomCurrencyDecimalSeparator = null;
         CustomCurrencyGroupSeparator = null;
         CustomCurrencyDecimalDigits = null;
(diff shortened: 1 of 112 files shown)


NickvisionMoney.Shared/Models/AccountMetadata.cs
```diff
@@ -121,6 +124,24 @@ internal AccountMetadata(string name, AccountType accountType)
...
+    /// The style to use for displaying an amount
+    /// </summary>
+    /// <remarks>Must be a value between 0 and 4. See https://learn.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.currencypositivepattern?view=net-7.0#remarks for the values' meaning</remarks>
+    public int? CustomCurrencyAmountStyle
+    {
+        get => _customCurrencyAmountStyle;
+
+        set
+        {
+            if (value != null && (value < 0 || value > 4))
```
fsobolev, line 137: `value > 3`?

fsobolev (changes requested):

fsobolev:
`n $` looks okay, but I think it can be improved. We could update ComboRow's model on changing currency symbol and use it, showing "100 🪙" for example.

NickvisionMoney.Shared/Models/AccountMetadata.cs
nlogozzo: Ah yes... could you change that and the comment for me?

nlogozzo:
> `n $` looks okay, but I think it can be improved. We could update ComboRow's model on changing currency symbol and use it, showing "100 🪙" for example.

That would work... world you like to tackle it or want me?

fsobolev:
> That would work... world you like to tackle it or want me?

I'm going to be busy for a while so I guess you will make it faster.

nlogozzo:
image
image

fsobolev (approved):

PR #676, opened 2023-08-29
Fix crash when transfering between accounts

### LINKED ISSUE #671: Crash when transfering money to another account
Everytime i want to transfer money to another account it's just crashed

```
UnhandledException - unhandled exception: System.ArgumentOutOfRangeException: Exception of type 'System.ArgumentOutOfRangeException' was thrown. (Parameter 'ordinal')
Actual value was 13.
   at Microsoft.Data.Sqlite.SqliteDataRecord.GetSqliteType(Int32 ordinal)
   at Microsoft.Data.Sqlite.SqliteDataRecord.IsDBNull(Int32 ordinal)
   at NickvisionMoney.Shared.Models.AccountMetadata.LoadFromAccountFile(String path, String password) in /run/build/org.nickvision.money/NickvisionMoney.Shared/Models/AccountMetadata.cs:line 196
   at NickvisionMoney.Shared.Controllers.TransferDialogController.UpdateTransfer(String destPath, String destPassword, String amountString, String sourceConversionAmountString, String destConversionAmountString) in /run/build/org.nickvision.money/NickvisionMoney.Shared/Controllers/TransferDialogController.cs:line 156
   at NickvisionMoney.GNOME.Views.TransferDialog.Validate() in /run/build/org.nickvision.money/NickvisionMoney.GNOME/Views/TransferDialog.cs:line 135
   at NickvisionMoney.GNOME.Views.TransferDialog.<.ctor>b__21_1(Object sender, NotifySignalArgs e) in /run/build/org.nickvision.money/NickvisionMoney.GNOME/Views/TransferDialog.cs:line 61
   at GObject.Closure.InternalCallback(IntPtr closure, IntPtr returnValuePtr, UInt32 nParamValues, ValueData[] paramValuesData, IntPtr invocationHint, IntPtr userData)
```

JoseBritto:
Closes #671

### DIFF
--- a/NickvisionMoney.GNOME/Views/AccountView.cs
+++ b/NickvisionMoney.GNOME/Views/AccountView.cs
@@ -56,6 +56,8 @@ public partial class AccountView : Adw.Bin
     private Dictionary<string, TagButton> _tagButtons;
     private Dictionary<uint, TransactionRow> _transactionRows;
     private uint _currentGraphPage;
+    private object _startupLock;
+    private bool _startupCalled;
 
     [Gtk.Connect] private readonly Adw.Flap _flap;
     [Gtk.Connect] private readonly Gtk.ScrolledWindow _paneScroll;
@@ -121,6 +123,8 @@ public partial class AccountView : Adw.Bin
         _tagButtons = new Dictionary<string, TagButton>();
         _transactionRows = new Dictionary<uint, TransactionRow>();
         _currentGraphPage = 0;
+        _startupLock = new object();
+        _startupCalled = false;
         //Register Controller Events
         _controller.AccountInformationChanged += (sender, e) => GLib.Functions.IdleAdd(0, AccountInformationChanged);
         _controller.GroupCreated += (sender, e) => GLib.Functions.IdleAdd(0, () => CreateGroupRow(e));
@@ -412,6 +416,11 @@ public partial class AccountView : Adw.Bin
     /// </summary>
     public async Task StartupAsync()
     {
+        lock (_startupLock)
+        {
+            if (_startupCalled) return;
+            _startupCalled = true;
+        }
         _paneScroll.SetSensitive(false);
         _viewStack.SetVisibleChildName("spinner");
         await Task.Run(async () => await _controller.StartupAsync());
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -27,6 +27,8 @@ public class AccountViewController : IDisposable
     private DateOnly _filterStartDate;
     private DateOnly _filterEndDate;
     private string _searchDescription;
+    private object _startupLock;
+    private bool _startupCalled;
 
     /// <summary>
     /// Gets the AppInfo object
@@ -188,6 +190,8 @@ public class AccountViewController : IDisposable
         _filteredExpense = 0;
         _groupFilters = new Dictionary<int, bool>();
         _tagFilters = new Dictionary<string, bool>();
+        _startupLock = new object();
+        _startupCalled = false;
         NotificationSent = notificationSent;
         RecentAccountsChanged = recentAccountsChanged;
         //Setup Filters
@@ -435,6 +439,11 @@ public class AccountViewController : IDisposable
     {
         if (!_isOpened)
         {
+            lock (_startupLock)
+            {
+                if (_startupCalled) return;
+                _startupCalled = true;
+            }
             await _account.LoadAsync();
             _searchDescription = "";
             //Metadata
--- a/NickvisionMoney.Shared/Controllers/AccountViewController.cs
+++ b/NickvisionMoney.Shared/Controllers/AccountViewController.cs
@@ -441,7 +441,12 @@ public class AccountViewController : IDisposable
         {
             lock (_startupLock)
             {
-                if (_startupCalled) return;
+                if (_startupCalled)
+                {
+                    while (!_isOpened)
+                        Task.Delay(100).Wait();
+                    return;
+                }
                 _startupCalled = true;
             }
             await _account.LoadAsync();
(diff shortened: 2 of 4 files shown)


JoseBritto:
Still getting this error. Not sure how to fix:
```
UnhandledException - unhandled exception: Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'UNIQUE constraint failed: transactions.id'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteNonQuery()
   at System.Data.Common.DbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at NickvisionMoney.Shared.Models.Account.AddTransactionAsync(Transaction transaction) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Models/Account.cs:line 850
   at NickvisionMoney.Shared.Models.Account.ReceiveTransferAsync(Transfer transfer, String description) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Models/Account.cs:line 1249
   at NickvisionMoney.Shared.Controllers.AccountViewController.ReceiveTransferAsync(Transfer transfer) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Controllers/AccountViewController.cs:line 909
   at NickvisionMoney.Shared.Controllers.MainWindowController.OnTransferSent(Object sender, Transfer transfer) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Controllers/MainWindowController.cs:line 387
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at GLib.Internal.MainLoopSynchronizationContext.<>c__DisplayClass1_0.<Post>b__0()
   at GLib.Internal.MainLoopSynchronizationContext.<>c__DisplayClass3_0.<ScheduleAction>b__0()


```

JoseBritto:
[9df73a7](https://github.com/NickvisionApps/Denaro/pull/676/commits/9df73a77c8e29388fe3b2ba24955b64e1e6d1481) might be bad code but I can't think of any other way to solve it. 

The previous error was also due to a race condition. Sometimes controller.ReceiveTransferAsync is called before Startup completes causing conflicts. This is probably why there was an explicit awaited call to controller.StartupAsync. Still this would cause problems as now there would be two Startup calls initializing the same account causing the errors I got first.

This async method isn't awaited properly:
https://github.com/NickvisionApps/Denaro/blob/4efc3fa134bce1de81ccd746a6d668ed767e3b93/NickvisionMoney.GNOME/Views/MainWindow.cs#L285-L298

But being an event I'm not sure there is a way to await it 🤔

fsobolev:
> But being an event I'm not sure there is a way to await it 🤔

```
AccountAdded += async (sender, e) => await AccountAdded();
```

should work.

> Still this would cause problems as now there would be two Startup calls initializing the same account causing the errors I got first.

Could you point to another place where StartupAsync is called? I think it's better to keep StartupAsync call in OnTransferSent and fix the other call instead of making the program slower as workaround.

JoseBritto:
Here: https://github.com/NickvisionApps/Denaro/blob/4efc3fa134bce1de81ccd746a6d668ed767e3b93/NickvisionMoney.GNOME/Views/MainWindow.cs#L297

But that can't be removed as it is responsible for initializing the account when its opened. Maybe the async call you will fix it. I'll try that

JoseBritto:
I tried this and removed the delay:
```
        _controller.AccountAdded += async (sender, e) => await AccountAdded(sender, e);
```
But still the same issue :(

Also made the method async Task instead of async void as void can't be awaited

JoseBritto:
> So await controller.StartupAsync will not return in OnTransferSent until startup is finished.

Ohh that should work.

fsobolev:
> `_controller.AccountAdded += async (sender, e) => await AccountAdded(sender, e);`

Actually... Now I wonder on what thread `AccountAdded` runs, it must be on the first one because it changes the UI.
And since it only calls async method in the end, I think you can try to make it sync.

```
_controller.AccountAdded += (sender, e) => Glib.IdleAdd(0, () => AccountAdded());
...
private void AccountAdded()
{
    ...
    Task.Run(newAccountView.StartupAsync());
}
```

I hope I don't miss something and this will work.

JoseBritto:
> Task.Run(newAccountView.StartupAsync());
But still that would cause it run on a seperate thread right? The problem we had was that when we don't know when exactly this one would complete so we can call the ReceiveTransfer

JoseBritto:
Also I used _isOpened variable for the check

fsobolev:
> But still that would cause it run on a seperate thread right?

Ah, I'm blind, thought it's the controller call 😅 Yeah, that doesn't solve the problem.

nlogozzo:
The whole lock system and everything isn't needed as that's what the `await` is for...it ensures the method is done before going to the next. As you can see from here: https://stackoverflow.com/questions/20084695/lock-and-async-method-in-c-sharp locks and async shouldn't be mixed...i didn't even know C# has locks haha

nlogozzo:
Going to open another PR with proper changes...

JoseBritto:
> The whole lock system and everything isn't needed as that's what the await is for

Minor correction, await waits for the async code (probably running on a separate thread) to finish. The current thread can do something else in the meantime. 

Locks are mainly for locking access to a variable so no other threads can touch it until the current thread is done with it.

https://stackoverflow.com/a/20084728

>lock is syntactic sugar for Monitor.Enter and Monitor.Exit. lock will keep execution in the method until the lock is released. It does not function like await in any way, shape or form.

PR #678, opened 2023-08-30
Fix Transfer Crash

### LINKED ISSUE #671: Crash when transfering money to another account

nlogozzo:
Closes #671
Fixes https://github.com/NickvisionApps/Denaro/issues/671#issuecomment-1696738448

(Replaces #676)

### DIFF
--- a/NickvisionMoney.Shared/Controllers/MainWindowController.cs
+++ b/NickvisionMoney.Shared/Controllers/MainWindowController.cs
@@ -383,7 +383,6 @@ public class MainWindowController : IDisposable
     {
         await AddAccountAsync(transfer.DestinationAccountPath, false, transfer.DestinationAccountPassword);
         var controller = _openAccounts.Find(x => x.AccountPath == transfer.DestinationAccountPath)!;
-        await controller.StartupAsync();
         await controller.ReceiveTransferAsync(transfer);
     }
 }
--- a/NickvisionMoney.Shared/Controllers/MainWindowController.cs
+++ b/NickvisionMoney.Shared/Controllers/MainWindowController.cs
@@ -378,12 +378,16 @@ public class MainWindowController : IDisposable
     /// <summary>
     /// Occurs when a transfer is sent from an account
     /// </summary>
+    /// <param name="sender">object?</param>
     /// <param name="transfer">The transfer sent</param>
     private async void OnTransferSent(object? sender, Transfer transfer)
     {
         await AddAccountAsync(transfer.DestinationAccountPath, false, transfer.DestinationAccountPassword);
-        await Task.Delay(500);
         var controller = _openAccounts.Find(x => x.AccountPath == transfer.DestinationAccountPath)!;
+        while (!controller.IsOpened)
+        {
+            await Task.Delay(200);
+        }
         await controller.ReceiveTransferAsync(transfer);
     }
 }
--- a/NickvisionMoney.Shared/Controllers/MainWindowController.cs
+++ b/NickvisionMoney.Shared/Controllers/MainWindowController.cs
@@ -382,6 +382,7 @@ public class MainWindowController : IDisposable
     private async void OnTransferSent(object? sender, Transfer transfer)
     {
         await AddAccountAsync(transfer.DestinationAccountPath, false, transfer.DestinationAccountPassword);
+        await Task.Delay(500);
         var controller = _openAccounts.Find(x => x.AccountPath == transfer.DestinationAccountPath)!;
         await controller.ReceiveTransferAsync(transfer);
     }
--- a/NickvisionMoney.Shared/Models/AccountMetadata.cs
+++ b/NickvisionMoney.Shared/Models/AccountMetadata.cs
@@ -141,39 +141,13 @@ public class AccountMetadata : ICloneable
             _customCurrencyAmountStyle = value;
         }
     }
-    
+
     /// <summary>
-    /// Loads metadata from an account file
+    /// Updates the metadata database table with new properties
     /// </summary>
-    /// <param name="path">The path to the account file</param>
-    /// <returns>AccountMetadata?</returns>
-    public static AccountMetadata? LoadFromAccountFile(string path, string? password)
+    /// <param name="database">SqliteConnection</param>
+    internal static void UpdateMetadataDatabaseTable(SqliteConnection database)
     {
-        if (Path.GetExtension(path).ToLower() != ".nmoney")
-        {
-            return null;
-        }
-        var connectionString = new SqliteConnectionStringBuilder()
-        {
-            DataSource = path,
-            Mode = SqliteOpenMode.ReadOnly,
-            Pooling = false
-        };
-        if (!string.IsNullOrEmpty(password))
-        {
-            connectionString.Password = password;
-        }
-        using var database = new SqliteConnection(connectionString.ConnectionString);
-        try
-        {
-            database.Open();
-        }
-        catch
-        {
-            database.Close();
-            return null;
-        }
-        // Update Metadata Table
         try
         {
             using var cmdTableMetadataUpdate1 = database.CreateCommand();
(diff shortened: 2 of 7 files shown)


nlogozzo:
@JoseBritto Try this PR should be good

JoseBritto:
Yea it should work. I'm not home rn so can't test. Will try once I get home. I added the startupCalled variable to prevent accidentally calling that method more than once in the future. The lock was there to prevent race conditions when accessing the variable. Its not part of the async code. As in if the method was to be called twice on two threads the method might actually end up running even if there was an if clause with a variable protecting it. 
Removing the second call works but I didn't find any other way to wait for initialization without adding arbitrary delay which looked bad to me :(

Didn't update the file from metadata because I'm not that good at sql😅 (also it will be updated anyway from somewhere else when the file is opened). 

Still think adding version info could be a good idea for the long term as it ensures we can change the way data is stored if needed in the future and not worry too much about breaking compatibility. 
On a similar note, if we abstract out the reading part from the upgrading the file part and upgrade only if needed, we can have a file open as readonly and if the user doesn't add any transactions, it wouldn't cause denaro to crash. I haven't tried to open a readonly file in denaro yet. Will try that too when I get home. 
Not sure if all of this would be worth the effort though haha. 

Let me know if you're interested in supporting opening readonly files. I'm not sure if it works as of now, but it is something I used to do with important files. But never tried with denaro.

JoseBritto:
Still crashing:
```
UnhandledException - unhandled exception: Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'UNIQUE constraint failed: transactions.id'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteNonQuery()
   at System.Data.Common.DbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at NickvisionMoney.Shared.Models.Account.AddTransactionAsync(Transaction transaction) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Models/Account.cs:line 850
   at NickvisionMoney.Shared.Models.Account.ReceiveTransferAsync(Transfer transfer, String description) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Models/Account.cs:line 1249
   at NickvisionMoney.Shared.Controllers.AccountViewController.ReceiveTransferAsync(Transfer transfer) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Controllers/AccountViewController.cs:line 900
   at NickvisionMoney.Shared.Controllers.MainWindowController.OnTransferSent(Object sender, Transfer transfer) in /home/britto/Source/C#/Denaro/NickvisionMoney.Shared/Controllers/MainWindowController.cs:line 387
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at GLib.Internal.MainLoopSynchronizationContext.<>c__DisplayClass1_0.<Post>b__0()
   at GLib.Internal.MainLoopSynchronizationContext.<>c__DisplayClass3_0.<ScheduleAction>b__0()
```

This is probably due to a race condition. I had seen this same one before. Its painful to debug these type of bugs. This one mostly happens (for me) when the app is in debug mode and transferring to this particular [file](0x0.st/Hp8B.nmoney).

JoseBritto:
Ha. Not a new bug. Seems like 500 isn't enough sometimes. This is the problem with hard coded delays. 1000 worked but it doesn't mean it will work everywhere tho.

NickvisionMoney.Shared/Controllers/MainWindowController.cs
```diff
@@ -382,8 +382,8 @@ public void RemoveRecentAccount(RecentAccount recentAccount)
     private async void OnTransferSent(object? sender, Transfer transfer)
     {
         await AddAccountAsync(transfer.DestinationAccountPath, false, transfer.DestinationAccountPassword);
+        await Task.Delay(500);
```
fsobolev, line 385: I remember you once complained about using delays as workarounds making the app slower :D And @JoseBritto confirmed my fear that 500 can be not enough. Maybe add something like `await controller.EnsureStartup();` that will not return until startup is finished (similar to what I suggested in previous PR)?

DaPigGuy (changes requested):

NickvisionMoney.Shared/Controllers/MainWindowController.cs
nlogozzo: done in latest commit.

NickvisionMoney.Shared/Models/AccountMetadata.cs
```diff
@@ -173,6 +173,56 @@ internal AccountMetadata(string name, AccountType accountType)
             database.Close();
             return null;
         }
+        // Update Metadata Table
+        try
```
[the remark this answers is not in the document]
nlogozzo: done.

nlogozzo:
@JoseBritto ok should be good now....I'm waiting for `IsOpened` to be `true` now instead of hardcoding a delay that way it can be as fast on small accounts and as slow on large accounts as it may need

nlogozzo:
> Still think adding version info could be a good idea for the long term as it ensures we can change the way data is stored if needed in the future and not worry too much about breaking compatibility.

Well instead of versioning i'd like to keep it as it is now and just update the tables itself with the changes as they come...reduces the code too...I moved updating the metadata table to it's own function that both `AccountMetadata` and `Account` can call to avoid duplication as @DaPigGuy suggested

> On a similar note, if we abstract out the reading part from the upgrading the file part and upgrade only if needed, we can have a file open as readonly and if the user doesn't add any transactions, it wouldn't cause denaro to crash. I haven't tried to open a readonly file in denaro yet. Will try that too when I get home.
Not sure if all of this would be worth the effort though haha.

Yeah I don't really think we should support `readonly` files as I see no point as Denaro will have no use just besides viewing a list of transactions, which I mean just export to a PDF or something it'll be cleaner formatted anyway haha

JoseBritto:
> If you mean checking "open file read-only" in the file chooser, it will not do anything, because we only use resulting GFile to get its path and then open the file using System.IO read-write anyway. And we always have read-write access in flatpak since we have home permission. And the reason for this is that I have no idea why but it seems SQL databases don't want to open through portal, I remember Denaro was crashing (this crash was fixed I think, it will just show that it can't open the file).

Oh wow. I didn't know that.

> Yeah I don't really think we should support readonly files as I see no point as Denaro will have no use just besides viewing a list of transactions, which I mean just export to a PDF or something it'll be cleaner formatted anyway haha

I meant exactly for viewing alone. Tbh, I prefer the UI of denaro lol. Also it allows me to filter by tags, see spending by groups, etc. That said, I don't use this kind of setup with denaro and it is probably a rare use case. Just wanted to throw it out there and see if that was something worth looking into. 

> Well instead of versioning i'd like to keep it as it is now and just update the tables itself with the changes as they come...reduces the code too...I moved updating the metadata table to it's own function that both AccountMetadata and Account can call to avoid duplication as @DaPigGuy suggested

Fair enough

JoseBritto:
> @JoseBritto ok should be good now....I'm waiting for `IsOpened` to be `true` now instead of hardcoding a delay that way it can be as fast on small accounts and as slow on large accounts as it may need


That seem to have fixed it.

PR #698, opened 2023-09-09
Read the rate from cache if available

UrtsiSantsi:
Before it was not read due to failing to create a DataTime from 'time_next_update_utc'. Cache is now written to disk only if parsing is successful. Refactor code by splitting into logical methods - reading, writing the rates from cache, getting rates from the service and parsing json are now in their own methods.

### DIFF
--- a/NickvisionMoney.Shared/Models/CurrencyConversionService.cs
+++ b/NickvisionMoney.Shared/Models/CurrencyConversionService.cs
@@ -89,55 +89,116 @@ public class CurrencyConversionService
     }
 
     /// <summary>
-    /// Gets a dictionary of conversion rates from the source currency
+    /// Gets a dictionary of conversion rates for the source currency
     /// </summary>
     /// <param name="sourceCurrency">The currency code to get converting rates for</param>
-    /// <returns>Dictionary&lt;string, decimal&gt; is successful, else false</returns>
+    /// <returns>Dictionary&lt;string, decimal&gt; is successful, else null</returns>
     /// <remarks>This method will cache the data for the sourceCurrency on disk</remarks>
     public static async Task<Dictionary<string, decimal>?> GetConversionRatesAsync(string sourceCurrency)
     {
-        var path = $"{UserDirectories.ApplicationCache}{Path.DirectorySeparatorChar}currency_{sourceCurrency}.json";
-        var needsUpdate = !File.Exists(path);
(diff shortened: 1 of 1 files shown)


nlogozzo (changes requested):
> Refactor code by splitting into logical methods - reading, writing the rates from cache, getting rates from the service and parsing json are now in their own methods.

There's really no reason for this as all these methods would only be called once by the main GetConversionRates method so we can just keep all the code that's there now.

Could you just apply the change of reading the unix time and making it a DateTime and just check that instead?

UrtsiSantsi:
> There's really no reason for this as all these methods would only be called once by the main GetConversionRates method
Following that logic we shouldn't have GetConversionRates since it is only called once by ConvertAsync ;). But seriously, I'm only adding a few lines (excluding doc comments and better logging) and now we have methods without side effects and we can easily change them separately.

Also, like I wrote in the commit message, cache is now written to disk only if parsing **conversion rates** is successful. Before it was cached even if it is invalid - for example due to some temporary issue with the service. The app will then be unable to get the rates after the service is fixed until the next day, since it will be finding a cache file. This is not hypothetical problem - I actually wrote an app that shows conversion rates with gir.core and I had several different temporary issues with the service (https://github.com/fawazahmed0/currency-api#readme), so when I saw in TWIG that you have the same functionality I decided to check what you use. Here is my app: 
image
Another issue is that if writing the cache file failed you didn't show conversion results, and I think it is more important to show rates, than to cache them.
I'm also checking that cache file exists only once :)

nlogozzo:
> Following that logic we shouldn't have GetConversionRates since it is only called once by ConvertAsync ;). But seriously, I'm only adding a few lines (excluding doc comments and better logging) and now we have methods without side effects and we can easily change them separately.

GetConversionRates() is called not just by ConvertAsync but in many places throughout controllers and views within the application. 
Method calls take up resources. We try to keep it at a minimum.

And thank you for implementing all those check and writing only when needed..like i said just keep it all in that one method and we will merge :)

UrtsiSantsi:
> Method calls take up resources. We try to keep it at a minimum.

It is not called in a loop so 2 additional method calls are nothing. Compared with better code readability it is worth it.
Anyway I created another merge request - #702 - with everything inside one method. You can compare the result and pick what you like.

PR #708, opened 2023-09-12
Shared.Tests - Add Tests

JoseBritto:
Tests added for CurrencyHelpers to make sure it works with all cultures.
 
 I hardcoded some values because in the tests we know exactly what values go into the test methods and what the expected output is.
 
 Currently all tests are passing except "kea-CV" and "pt-CV". These cultures have $ in the middle of the number: 
 $1.00 -> 1$00 
 $2.05 -> 2$05
 Also for some reason they also have a zero width space at the end of the number.

### DIFF
--- a/NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
+++ b/NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
@@ -1,0 +1,120 @@
+using System.Globalization;
+using NickvisionMoney.Shared.Helpers;
+using Xunit;
+
+namespace NickvisionMoney.Shared.Tests;
+
+public class CurrencyHelperTests
+{
+    public static decimal[] SampleAmounts = { 0M, 109M, 100M, 10920M, 0.002M, 1.2M, 12.00000004M, 1.0234567890M };
+
+    public static IEnumerable<object[]> GetSampleData()
+    {
+        var cultures = CultureInfo.GetCultures(CultureTypes.SpecificCultures).AsEnumerable();
+        //Add a custom culture
+        var goldCulture = CultureInfo.InvariantCulture.Clone() as CultureInfo;
+        goldCulture!.NumberFormat.CurrencyDecimalDigits = 99;
+        goldCulture.NumberFormat.CurrencyDecimalSeparator = "*";
+        goldCulture.NumberFormat.CurrencyGroupSeparator = "/";
+        goldCulture.NumberFormat.CurrencySymbol = "\ud83e\ude99";
+        cultures = cultures.Append(goldCulture);
+        foreach (var culture in cultures)
+        {
+            foreach (var number in SampleAmounts)
+            {
+                yield return new object[] { culture, number };
+            }
+        }
+    }
+    
+    [Theory]
+    [MemberData(nameof(GetSampleData))]
+    public void ToAmountString_AllCulturesShouldWorkByDefault(CultureInfo culture, decimal amount)
+    {
+        var expected = amount.ToString("C", culture);
+        if (culture.NumberFormat.CurrencyDecimalDigits == 99 && expected.Contains(culture.NumberFormat.CurrencyDecimalSeparator))
+        {
+            expected = expected.TrimEnd('0');
+            if (expected.EndsWith(culture.NumberFormat.CurrencyDecimalSeparator))
+                expected = expected.Replace(culture.NumberFormat.CurrencyDecimalSeparator, "");
+        }
+        var result = amount.ToAmountString(culture, false);
+        Assert.Equal(expected, result);
+    }
+    
+    [Theory]
+    [MemberData(nameof(GetSampleData))]
+    public void ToAmountString_AllCulturesShouldWorkWithNativeDigits(CultureInfo culture, decimal amount)
+    {
+        var expected = amount.ToString("C", culture);
+        if (culture.NumberFormat.CurrencyDecimalDigits == 99 && expected.Contains(culture.NumberFormat.CurrencyDecimalSeparator))
+        {
+            expected = expected.TrimEnd('0');
+            if (expected.EndsWith(culture.NumberFormat.CurrencyDecimalSeparator))
+                expected = expected.Replace(culture.NumberFormat.CurrencyDecimalSeparator, "");
+        }
+        expected =  expected
+            .Replace("0", culture.NumberFormat.NativeDigits[0])
+            .Replace("1", culture.NumberFormat.NativeDigits[1])
+            .Replace("2", culture.NumberFormat.NativeDigits[2])
+            .Replace("3", culture.NumberFormat.NativeDigits[3])
+            .Replace("4", culture.NumberFormat.NativeDigits[4])
+            .Replace("5", culture.NumberFormat.NativeDigits[5])
+            .Replace("6", culture.NumberFormat.NativeDigits[6])
+            .Replace("7", culture.NumberFormat.NativeDigits[7])
+            .Replace("8", culture.NumberFormat.NativeDigits[8])
+            .Replace("9", culture.NumberFormat.NativeDigits[9]);
+        var result = amount.ToAmountString(culture, true);
+        Assert.Equal(expected, result);
+    }
+    
+    [Theory]
+    [MemberData(nameof(GetSampleData))]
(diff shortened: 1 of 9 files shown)


fsobolev:
> Currently all tests are passing except "kea-CV" and "pt-CV". These cultures have $ in the middle of the number:
$1.00 -> 1$00
$2.05 -> 2$05
Also for some reason they also have a zero width space at the end of the number.

The values seem to be correct: these cultures don't have currency sign (that probably explains space at the end) and $ as decimal separator.
[kea-CV](https://www.localeplanet.com/icu/kea-CV/index.html) [pt-CV](https://www.localeplanet.com/icu/pt-CV/index.html)

JoseBritto:
Those are the expected values. Currently those tests fail.

fsobolev:
> Those are the expected values. Currently those tests fail.

Ah okay, got it. Let's see...

JoseBritto:
Assert.Equal() Failure
          ↓ (pos 0)
Expected: 0$00 ​
Actual:   $00

Wasn't sure what the best way to solve it is, so I didnt touch the method itself

NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
```diff
@@ -0,0 +1,194 @@
...
+    public void ToAmountString_AllCulturesShouldWorkWithOverwriteDecimal(CultureInfo culture, decimal amount)
+    {
+        //Arrange
+        var expected = amount.ToString("C6", culture).Replace("\u200b", "").Trim();
+        if (culture.Name is "kea-CV" or "pt-CV")
+        {
+            expected = expected.TrimEnd('0');
+            if (expected.EndsWith(culture.NumberFormat.CurrencySymbol))
+                expected += "00"; // Make the currency symbol be in the center of the string
+        }
```
JoseBritto, line 139: Not sure what is supposed to be done here 🤔 Is it ok to leave the currency at the end? If not how many zeros should be added

JoseBritto: For context: this is in a special condition for the two cultures that use $ instead of .

JoseBritto:
TODO: 
1. Fix ToAmountString to account for $ instead of decimal point
2. Add `dotnet test` to github workflows? (so we can have future tests automatically run with the other checks)

Not sure how to do either of them at the moment.

fsobolev:
> Assert.Equal() Failure ↓ (pos 0) Expected: 0$00 ​Actual: $00
> 
> Wasn't sure what the best way to solve it is, so I didnt touch the method itself

I fixed the method to pass the tests that was giving results like this, but there are also the tests that give the opposite:

```
Assert.Equal() Failure
          ↓ (pos 0)
Expected: $02 ​
Actual:   1$02 ​
          ↑ (pos 0)
```

NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
fsobolev: A lot of cultures use symbols other than `.` as decimal separator, I don't understand why do you see a problem with `$`, it should be treated like any other decimal separator. As I understand all the mess is because of invisible currency symbol. 🤔

nlogozzo:
> Add dotnet test to github workflows? (so we can have future tests automatically run with the other checks)

I'll make the workflow for this

nlogozzo:
https://github.com/NickvisionApps/Denaro/actions/runs/6158369998

NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
JoseBritto: Nono. $ is not just the decimal separator here, it is the currency symbol itself. So if we have unlimited decimals and the input is 2$00 according to our previous logic we trim all zeros and remove the separator. But we can't remove the separator because it is the currency symbol too, so 2$ it is. Now another problem, the $ is no  longer at the middle but instead at the end. What i did here is add back 00 in this case and make it 2$00. Not sure if this should be the expected behavior.

JoseBritto (commented):
Are we fixing the method separately? This one just makes the test pass but doesn't actually fix the problem 🤔 (about the tests)

JoseBritto (commented):
https://github.com/NickvisionApps/Denaro/pull/708#pullrequestreview-1622010830
I meant to leave that comment here. AAh that did the same thing 🤦
Meant to comment on Fyodor's commits

NickvisionMoney.Shared.Tests/CurrencyHelperTests.cs
fsobolev: Wow, that's complex, sorry, I think I understand now, gonna undo my commit and see what can be done

fsobolev:
@JoseBritto let me sumarize info about problematic cultures to ensure I don't miss anything:
1. `CurrencySymbol` - `\u200b` (non-printable)
2. `CurrencyDecimalSeparator` - `$`
3. Format can't be C, it should be minimum C2 (unless `overwriteDecimal` is true), so 1.0 becomes 1$00
4. Currency symbol seem to cause misbehavior of `string.Remove`, `string.Replace` should be used instead

If all this is correct, I have a fix for the method itself, but the tests are complaining about symbol in the end (I see from errors reports that other than that the values are correct)

JoseBritto:
> @JoseBritto let me sumarize info about problematic cultures to ensure I don't miss anything:
> 
>     1. `CurrencySymbol` - `\u200b` (non-printable)
> 
>     2. `CurrencyDecimalSeparator` - `$`
> 
>     3. Format can't be C, it should be minimum C2 (unless `overwriteDecimal` is true), so 1.0 becomes 1$00
> 
>     4. Currency symbol seem to cause misbehavior of `string.Remove`, `string.Replace` should be used instead
> 
> 
> If all this is correct, I have a fix for the method itself, but the tests are complaining about symbol in the end (I see from errors reports that other than that the values are correct)

Yes. I didn't actually care to check what the currency symbol was actually returning, assumed it was $ as that is what google told me and got confused when I got the invisible character at the end 🤦

fsobolev:
> 3\. Format can't be C, it should be minimum C2 (unless `overwriteDecimal` is true), so 1.0 becomes 1$00

Just noticed 1.2 was becoming 1$2 in AllCulturesShouldWorkWithOverwriteDecimal test, fixed actual value in latest commit.

fsobolev:
Just one test left, this time it fails with mm-Mong-MN culture:

```
Assert.Equal() Failure
             ↓ (pos 3)
Expected: ₮10,920
Actual:   ₮10920
             ↑ (pos 3)
```

No idea why 🤔

JoseBritto:
I managed to get all tests to pass and make the tests a lot simpler i think. Will pull in https://github.com/NickvisionApps/Denaro/pull/708/commits/4f7d2d0dada246c429df91595caba6d1a3b67918 and see if it works with that. If it does can i revert your latest commit? @fsobolev

fsobolev:
> Will pull in [4f7d2d0](https://github.com/NickvisionApps/Denaro/commit/4f7d2d0dada246c429df91595caba6d1a3b67918) and see if it works with that. If it does can i revert your latest commit? @fsobolev

Sure, feel free to do whatever you need :)

JoseBritto:
Didn't have to revert. I'm bad at git lol. Anyway, now it should look better than the mess I made yesterday

nlogozzo:
LGTM! Any remaining issues?

JoseBritto:
Not that I know of. I'll wait or @fsobolev just in case

nlogozzo (approved):

fsobolev (approved):

fsobolev:
Ouch, I didn't know rebase adds person who make a rebase as co-author, I hope nobody minds

JoseBritto:
> Ouch, I didn't know rebase adds person who make a rebase as co-author (at least when using Github Desktop), I hope nobody minds

I don't mind lol. This should now be good to merge right?

fsobolev:
> This should now be good to merge right?

Yes, I can't merge because I'm the last pusher

nlogozzo:
LGTM

anaximeno:
Thanks for implementing the "kea-CV" and "pt-CV", I was here to check something related to this, but seen you've already implemented it, so thanks again.

PR #716, opened 2023-09-21
Refactor Account

### LINKED ISSUE #715: Bug Report - Incorrect Values in PDF Export from Current View
Description:
I would like to report a bug regarding the PDF export feature in our application. When exporting the current view to PDF format, the overview values (total expenses and total income) do not reflect the data from the current view; instead, they display the values from the whole account.

Expected Behavior:
The expected behaviour is that when exporting the current view to PDF, the overview values (total expenses and total income) should accurately represent the data present in the current view.

Steps to Reproduce:

1. Create a view, for instance, by clicking on the button labelled **Select current month**.
2. Click on the "Actions" menu and select the "Export as PDF" option.
3. Choose the "Export current view" option in the PDF export dialogue.
4. Observe that the generated PDF contains incorrect values for total expenses and total income, which should match the data in the current view.

Version: 2023.9.2

JoseBritto:
Fixes #715

The overview and Groups section showed incorrect values when exported to pdf using 'Export Current View'. Now it matches what's shown in the app.

Moved helper methods into AccountHelpers

I think I saw some repeated logic somewhere, need to look those up make them use these methods instead. If we keep the helpers separated we can have them covered by tests easily. 

I tried my best to split everything into small commits to make reviewing changes easier. 
@nlogozzo @fsobolev have a look and see if these changes are desirable. If not I'll close this pr and open a new one without AccountHelpers.


nlogozzo:
> If we keep the helpers separated we can have them covered by tests easily.

So the whole reason for `AccountHelpers` is to make tests easier, correct? Why can't we just use the `Account` class directly?

JoseBritto:
Account class has too many dependencies. Like loading from disk and stuff and its hard to mock those to test.

JoseBritto:
Yea I guess IO is the only problem with testing it.

JoseBritto:
Also doesn't this improve readability? That was my main motive here tbh. Almost 300 or so lines were removed from the Account class.

nlogozzo:
> Account class has too many dependencies. Like loading from disk and stuff and its hard to mock those to test.

I only see loading the file as the only dependency here...is it really that hard to create a test file when running tests and deleting it once they're over? Plus I feel like having a test file would test our database infrastructure as well and catch any sql errors...

> Also doesn't this improve readability? That was my main motive here tbh. Almost 300 or so lines were removed from the Account class.

IMO, no. This is creating more fragmentation removing critical functionality away from the Account such as getting income, expense, and totals and creating graphs.

I mean the only refactor I could maybe be okay with is moving the database code out into a AccountStore class in case we ever need to switch SQL to another database sceme...but I rejected that once cause I never saw us switching to another other than SQL (and i still have no plans of doing so) but if Account is really that big for everyone, that'd probably be the only refactor I'll reconsider

JoseBritto:
> I only see loading the file as the only dependency here...is it really that hard to create a test file when running tests and deleting it once they're over? Plus I feel like having a test file would test our database infrastructure as well and catch any sql errors...

That is integration tests which is different from unit tests. The goal of the unit tests is to have as few dependcies as possible and just test the logic.
Testing with an actual file is needed imo but I couldn't think of any good way to do that.

> IMO, no. This is creating more fragmentation removing critical functionality away from the Account such as getting income, expense, and totals and creating graphs.

Hmmm...

Are the refectors good with you if they are in the same file they were? If so I'll do that then

JoseBritto:
Or its probably better to start it in a new pr. This one already has too many commits. I dont want to pollute the repo with unnecessary commits. I'll close this and create a new pr with only the bug fix tomorrow. From there we can refactor if needed instead of going too far and then having to revert.

JoseBritto:
Reopen if you have some other plan. Its pretty late for me. I'll create the new pr tomorrow.

PR #717, opened 2023-09-22
Fix PDF Export bug

### LINKED ISSUE #715: Bug Report - Incorrect Values in PDF Export from Current View

JoseBritto:
Fixes #715 
Replaces #716  

- Added GetGroupIncome,Expense and Total methods.
- Simplified existing GetIncome,Expense and Total methods

@nlogozzo Let me know if anything needs to be changed.


nlogozzo:
Could you update the changelog in `GNOME/Program.cs` and `Shared/org.nickvision.money.metainfo.xml`? 

Something like `- Fixed an issue where exported PDF values were incorrect` is fine.

NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -550,90 +550,94 @@ public async Task<bool> LoadAsync()
...
-    /// Gets the income amount for the date range
+    /// Gets the total income amount for the transactions given
     /// </summary>
-    /// <param name="endDate">The end date</param>
-    /// <param name="startDate">The start date</param>
-    /// <returns>The income amount for the date range</returns>
-    public decimal GetIncome(DateOnly endDate, DateOnly? startDate = null)
+    /// <param name="transactionIds">The ids of transactions to consider</param>
+    /// <returns>The income amount</returns>
+    public decimal GetIncome(IEnumerable<uint>? transactionIds = null)
```
nlogozzo, line 557: You can simplify these even more to just:
```csharp
public decimal GetIncome(IEnumerable<uint>? transactionIds = null) => (transactionIds ?? Transactions.Keys)
            .Select(id => Transactions[id])
            .Where(transaction => transaction.Type == TransactionType.Income)
            .Sum(transaction => transaction.Amount);
```
and the other methods as well.

nlogozzo (changes requested):

NickvisionMoney.Shared/Models/Account.cs
JoseBritto: Done. Thanks

JoseBritto:
I'll refactor the generate graphs to use these rn

nlogozzo (approved):

nlogozzo:
Thank you so much for this :)

PR #735, opened 2023-11-15
Use IsNullOrWhitespace instead of IsNullOrEmpty

### LINKED ISSUE #730: Invisible groups with whitespace name can be created
The groups exist and another one with the same name cannot be create, but they are not visible. And they are also meaningless.

UrtsiSantsi:
Decimal and group separators are not touched - as discussed they can be white space.
Currency name and symbol are also left untouched, but I think they cannot be white space.

Fixes #730

### DIFF
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -431,8 +431,8 @@ public class Account : IDisposable
             Metadata.Name = readQueryMetadata.GetString(1);
             Metadata.AccountType = (AccountType)readQueryMetadata.GetInt32(2);
             Metadata.UseCustomCurrency = readQueryMetadata.GetBoolean(3);
-            Metadata.CustomCurrencySymbol = string.IsNullOrEmpty(readQueryMetadata.GetString(4)) ? null : readQueryMetadata.GetString(4);
-            Metadata.CustomCurrencyCode = string.IsNullOrEmpty(readQueryMetadata.GetString(5)) ? null : readQueryMetadata.GetString(5);
+            Metadata.CustomCurrencySymbol = string.IsNullOrWhiteSpace(readQueryMetadata.GetString(4)) ? null : readQueryMetadata.GetString(4);
+            Metadata.CustomCurrencyCode = string.IsNullOrWhiteSpace(readQueryMetadata.GetString(5)) ? null : readQueryMetadata.GetString(5);
             Metadata.DefaultTransactionType = (TransactionType)readQueryMetadata.GetInt32(6);
             Metadata.ShowGroupsList = readQueryMetadata.GetBoolean(7);
             Metadata.SortFirstToLast = readQueryMetadata.GetBoolean(8);
--- a/NickvisionMoney.Shared/Models/Account.cs
+++ b/NickvisionMoney.Shared/Models/Account.cs
@@ -176,7 +176,7 @@ public class Account : IDisposable
         set
         {
             //Remove Password If Empty (Decrypts)
-            if (string.IsNullOrEmpty(value))
+            if (string.IsNullOrWhiteSpace(value))
             {
                 //Create Temp Decrypted Database
                 var tempPath = $"{Path}.decrypt";
@@ -308,7 +308,7 @@ public class Account : IDisposable
             //Set Password
             if (IsEncrypted)
             {
-                if (string.IsNullOrEmpty(password))
+                if (string.IsNullOrWhiteSpace(password))
                 {
                     _loggedIn = false;
                     return false;
@@ -514,13 +514,13 @@ public class Account : IDisposable
                 RGBA = readQueryTransactions.IsDBNull(7) ? "" : readQueryTransactions.GetString(7),
                 UseGroupColor = readQueryTransactions.IsDBNull(11) ? false : readQueryTransactions.GetBoolean(11),
                 RepeatFrom = readQueryTransactions.IsDBNull(9) ? -1 : readQueryTransactions.GetInt32(9),
-                RepeatEndDate = readQueryTransactions.IsDBNull(10) ? null : (string.IsNullOrEmpty(readQueryTransactions.GetString(10)) ? null : DateOnly.Parse(readQueryTransactions.GetString(10), new CultureInfo("en-US", false))),
+                RepeatEndDate = readQueryTransactions.IsDBNull(10) ? null : (string.IsNullOrWhiteSpace(readQueryTransactions.GetString(10)) ? null : DateOnly.Parse(readQueryTransactions.GetString(10), new CultureInfo("en-US", false))),
                 Notes = readQueryTransactions.IsDBNull(12) ? "" : readQueryTransactions.GetString(12),
-                Tags = readQueryTransactions.IsDBNull(13) ? new List<string>() : (string.IsNullOrEmpty(readQueryTransactions.GetString(13)) ? new List<string>() : readQueryTransactions.GetString(13).Split(',').ToList())
+                Tags = readQueryTransactions.IsDBNull(13) ? new List<string>() : (string.IsNullOrWhiteSpace(readQueryTransactions.GetString(13)) ? new List<string>() : readQueryTransactions.GetString(13).Split(',').ToList())
             };
             Tags.AddRange(transaction.Tags.Where(t => !Tags.Contains(t)));
             var receiptString = readQueryTransactions.IsDBNull(8) ? "" : readQueryTransactions.GetString(8);
-            if (!string.IsNullOrEmpty(receiptString))
+            if (!string.IsNullOrWhiteSpace(receiptString))
             {
                 transaction.Receipt = SixLabors.ImageSharp.Image.Load(Convert.FromBase64String(receiptString));
             }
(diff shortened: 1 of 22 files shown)


nlogozzo:
This needs to be rebased with `main` as some of these files don't exist anymore, i.e. `NickvisionMoney.GNOME/Helpers/Help.cs`

nlogozzo:
CC @JoseBritto

JoseBritto:
User provided custom currencies can be empty. Right now you can create an account with currency name and symbol as whitespace. Checking for whitespace here should do the trick.
https://github.com/NickvisionApps/Denaro/blob/111308064333cfa9712b6bf73a53739b1e54a3d0/NickvisionMoney.Shared/Controllers/NewAccountDialogController.cs#L123-L149

Line 150 shouldn't be changed as if its changed to nullorwhitespace it would allow you to have to space in currency symbol and use space as the group separator, which isn't allowed I believe. Nick correct me if im wrong on this one.

nlogozzo:
@UrtsiSantsi will you be able to finish this by Thursday, the latest?

If not, @JoseBritto could you make that change and make sure it all looks good?

I would like to release a stable version on Friday with this included.

JoseBritto:
@nlogozzo tbh, I'm not sure how to modify this pr 😅. I'll have to look up some guides from github.

Also one more small thing. 
I think we should trim user inputs in some places. Right now a user can name an account a<space> and it will be created as 'a .nmoney' which can cause some confusions. Most other applications would just remove the space

JoseBritto:
@UrtsiSantsi If you're still having trouble with the environment let me know. I'll push the changes. Also maybe I can help with setting it up.

nlogozzo:
Currency name and symbol should not be whitespace, I agree with you guys

JoseBritto:
>  I think we should trim user inputs in some places. Right now a user can name an account a<space> and it will be created as 'a .nmoney' which can cause some confusions. Most other applications would just remove the space

The same thing happens with groups. There can be two groups one 'abc\<space>' and another 'abc' without the space and denaro will treat them differently. Tags are not affected by this as they are already trimed with the GetText() here:
https://github.com/NickvisionApps/Denaro/blob/58ed6ab35acefec3ca1b2ac1298514785b53fd04/NickvisionMoney.GNOME/Views/TransactionDialog.cs#L353

Replacing GetText() with GetText().Trim() here
https://github.com/NickvisionApps/Denaro/blob/58ed6ab35acefec3ca1b2ac1298514785b53fd04/NickvisionMoney.GNOME/Views/GroupDialog.cs#L143
and here
https://github.com/NickvisionApps/Denaro/blob/58ed6ab35acefec3ca1b2ac1298514785b53fd04/NickvisionMoney.GNOME/Views/NewAccountDialog.cs#L288
would fix group names and account names.

@UrtsiSantsi can you accommodate that as well along with the other fix. I don't think there is a need to create a separate pr for this.

JoseBritto:
I'll try it out

NickvisionMoney.Shared/Models/Account.cs
```diff
@@ -308,7 +308,7 @@ public bool Login(string? password)
             //Set Password
             if (IsEncrypted)
             {
-                if (string.IsNullOrEmpty(password))
+                if (string.IsNullOrWhiteSpace(password))
```
JoseBritto, line 311: Didn't catch this earlier. This one would prevent accessing accounts stored with an older version which has its password set as a whitespace. I don't why would anyone do that but since we allowed it earlier, someone out there could have done it

JoseBritto:
With commit [bd1fa10](https://github.com/NickvisionApps/Denaro/pull/735/commits/bd1fa102552b9c2a6a465868f179d6a39f5ebfee) previously saved accounts with empty currency symbol will be now shown as the local symbol. I'm not sure if thats desirable or not. In my opinion we are probably better off not checking for whitespaces when loading data from nmoney files as the data is written by denaro after validation anyway. 

Ignoring passwords which are just white spaces feels a bit off since they show up as characters in the password field in the new account dialog but the account gets created without a password. I feel like we should leave that as it was earlier.

@nlogozzo Thoughts?

JoseBritto:
Not sure why snap is failing 🤔

JoseBritto:
Writing this here as I always forget this 😅 We should update the changelog as well

NickvisionMoney.Shared/Models/Account.cs
nlogozzo: Yes I'd revert this back to `string.IsNullOrEmpty()` for backward compatibility

nlogozzo:
> Not sure why snap is failing 🤔

snap has an old version of `blueprint-compiler` that doesn't support `Adw.Breakpoint` syntax. Don't worry about it @soumyaDghosh is on the way with a fix

nlogozzo:
> With commit [bd1fa10](https://github.com/NickvisionApps/Denaro/pull/735/commits/bd1fa102552b9c2a6a465868f179d6a39f5ebfee) previously saved accounts with empty currency symbol will be now shown as the local symbol. I'm not sure if thats desirable or not. In my opinion we are probably better off not checking for whitespaces when loading data from nmoney files as the data is written by denaro after validation anyway.

I think it's okay because then users will notice and they will be forced into going into AccountSettings to fix their symbol and not provide a whitespace, to match the new standards.

> Ignoring passwords which are just white spaces feels a bit off since they show up as characters in the password field in the new account dialog but the account gets created without a password. I feel like we should leave that as it was earlier. Or is there a way to hint the user that the password is invalid?

Yeah I'd say passwords can be whitespace cause ` ` vs `  ` vs `   ` could all be valid passwords is my opinion but they would fail validation/login.

nlogozzo:
> but it will fail validation, so I think this check should stay.

We are saying to revert it back to IsNullOrEmpty for all passwords checks because we want the different whitespace to all be valid.

JoseBritto:
> > Yeah I'd say passwords can be whitespace cause ` ` vs ` ` vs ` ` could all be valid passwords is my opinion but they would fail validation/login.
> 
> Yes, you can construct password that consist only of different types of white space, but it will fail validation, so I think this check should stay.

The problem right now is that when typing in just spaces the password prompt shows input just like any other character and the user believes that a password is being set correctly but actually since the whitespace password is detected as empty (even though the user was shown the password field containing characters) denaro creates a regular unencrypted account. There is currently nothing informing the user that its not a valid password. 

image 
*I typed in two spaces here*

To disallow whitespace passwords properly, we need to have some kind of ruleset for password validation and inform users before they move on from this screen. Then again whitespace should be considered *invalid* password instead of *empty* password. Treating whitespace as the same as an empty password field would cause confusion.

NickvisionMoney.Shared/Models/Account.cs
JoseBritto: > While it is possible, it is quite unlikely. Lets leave it like this and if somebody complains, we can restore it.

Imo, that is not a good approach. Many people, even though on linux don't know how to create an issue 😅. I agree that is a very rare use case though.

nlogozzo: When though it is a very rare case, I believe we shouldn't throw those users away and break it for them. `string.IsNullOrEmpty` it should stay...

As mentioned below, we want to keep whitespace passwords anyway so

nlogozzo:
> PS: I'm not saying that we should disallow whitespace passwords, but incase we do it, it should be done properly

I'm against disallowing whitespace passwords. I think they should stay.

JoseBritto:
> I'm against disallowing whitespace passwords. I think they should stay.
Everything else expect the parts where it touches passwords should be good then

nlogozzo:
> > I'm against disallowing whitespace passwords. I think they should stay.
> 
> Everything else expect the parts where it touches passwords should be good then

To you have time now? Could you update this PR and restore the password stuff back to `IsNullOrEmpty` so I can merge in the morning and release stable?

JoseBritto:
Sure thing

nlogozzo:
> Sure thing

Thank you :)

JoseBritto:
@nlogozzo What do you think about including [this](https://github.com/NickvisionApps/Denaro/pull/735#issuecomment-1813025308) in this release as well? It seems like a minor thing fix to me.

nlogozzo:
> @nlogozzo What do you think about including [this](https://github.com/NickvisionApps/Denaro/pull/735#issuecomment-1813025308) in this release as well? It seems like a minor thing fix to me.

We can include that. Feel free to commit that all in here too so it's just one merge.

NickvisionMoney.Shared/Models/Account.cs
JoseBritto: reverted

JoseBritto:
Not sure if the changelog is good enough. You might have to rephrase it tomorrow, I'm bad at writing these things 😅

nlogozzo:
> Not sure if the changelog is good enough. You might have to rephrase it tomorrow, I'm bad at writing these things 😅

Changelog is fine. It'd just but the disallowed as the first item and keep all the fixes together. But I'll edit that myself.

Thanks so much for working on this, you and @UrtsiSantsi ❤️

nlogozzo (approved):

PR #741, opened 2023-11-17
build `blueprint-compiler` from source

soumyaDghosh:
Changes in this PR

1. Building `blueprint-compiler` from source
2. removing the patch

@fsobolev do you think that the patch worth maintaining?


nlogozzo:
> do you think that the patch worth maintaining?

What was the purpose of the patch in the first place? Wasn't it because the icon didn't show without it?

soumyaDghosh:
> What was the purpose of the patch in the first place? Wasn't it because the icon didn't show without it?

The patch is there to show the symbolic icon in the gnome activities icon.

nlogozzo:
> > What was the purpose of the patch in the first place? Wasn't it because the icon didn't show without it?
> 
> The patch is there to show the symbolic icon in the gnome activities icon.

So why would we get rid of that...

soumyaDghosh:
> > > What was the purpose of the patch in the first place? Wasn't it because the icon didn't show without it?
> > 
> > 
> > The patch is there to show the symbolic icon in the gnome activities icon.
> 
> So why would we get rid of that...

This much change just for a symbolic icon! It's not like the app doesn't show any icon, it just shows the black and white version of the app's original icon

nlogozzo:
> This much change just for a symbolic icon!

I don't mind maintaining it to show the correct icon. It's not like the desktop file changes often where we need to fix the patch. Just when we move things around, but I update the patch.

Please restore it back, I'll maintain it dw

snap/denaro.patch
nlogozzo, line 1: Restore please

.github/workflows/snap.yml
```diff
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "main", "snap" ]
```
nlogozzo, line 3: remove `"snap"` as testing is over

snap/snapcraft.yaml
```diff
@@ -1,32 +1,46 @@
...
+    #build-snaps: # when available
+    #- blueprint-compiler
+    build-packages: 
       - wget
     override-pull: |
       craftctl default
       wget https://dot.net/v1/dotnet-install.sh -O dotnet-install.sh
       chmod +x ./dotnet-install.sh
       ./dotnet-install.sh --channel 8.0
-      patch -p1 < $CRAFT_PROJECT_DIR/snap/denaro.patch
```
nlogozzo, line 26: restore

nlogozzo (changes requested):

nlogozzo (approved):

nlogozzo:
Could you do the same PR for Tagger please?

soumyaDghosh:
Sure.. just a moment, working on `snapcraft`

Issue #762, opened 2023-12-24
Nothing to import from the file


bennypowers:
I've generated .ofx exports from [toshl finance](https://toshl.com). Here's a redacted minimal repro:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?OFX OFXHEADER="200" VERSION="220" SECURITY="NONE" OLDFILEUID="NONE" NEWFILEUID="NONE"?>
<OFX>
    <BANKMSGSRSV1>
        <STMTTRNRS>
            <CURDEF>ILS</CURDEF>
            <BANKACCTFROM>
                <BANKID>000</BANKID>
                <ACCTID>000</ACCTID>
                <ACCTTYPE>CHECKING</ACCTTYPE>
            </BANKACCTFROM>
            <BANKTRANLIST>
                <DTSTART>20130615000000.000[+0000]</DTSTART>
                <DTEND>20231220000000.000[+0000]</DTEND>
                <STMTTRN>
                    <TRNTYPE>OTHER</TRNTYPE>
                    <DTPOSTED>20130501000000.000[+0000]</DTPOSTED>
                    <TRNAMT>-47.00</TRNAMT>
                    <FITID>17100-0</FITID>
                    <MEMO>Kupat Holim</MEMO>
                    <SIC>🏥 Health &amp; Personal Care;Health Insurance 💳, insurance</SIC>
                </STMTTRN>
            </BANKTRANLIST>
        </STMTTRNRS>
    </BANKMSGSRSV1>
</OFX>
```
If I remove the emoji and hebrew glyphs, and the tags from the `<SIC>` tag (the `;` and everything to the right), it's the same result. Same thing if I change the `<CURDEF>` to `USD`

JoseBritto:
I'm not sure what is there to be signed into here but the exception is thrown is this:
  
```
 OfxSharp.OFXParseException: Sign On information not found
   at OfxSharp.OFXDocumentParser.Parse(String ofxString)
   at OfxSharp.OFXDocumentParser.ParseOfxDocument(String ofxString)
   at OfxSharp.OFXDocumentParser.Import(String ofx)
   at NickvisionMoney.Shared.Models.Account.ImportFromOFXAsync(String path, String defaultTransactionRGBA) 
  
  NickvisionMoney.Shared/Models/Account.cs:line 1428
  ```
  At this line:
[...]

PR #763, opened 2023-12-25
Print to stderr when there is an exception

JoseBritto:
This PR should add logs in missed places.

### DIFF
--- a/NickvisionMoney.GNOME/Views/AccountView.cs
+++ b/NickvisionMoney.GNOME/Views/AccountView.cs
@@ -808,7 +808,7 @@ public partial class AccountView : Adw.BreakpointBin
             }
             _controller.ExportToCSV(path ?? "", exportMode);
         }
-        catch(Exception e)
+        catch (Exception e)
         {
             Console.Error.WriteLine(e);
         }
--- a/NickvisionMoney.GNOME/Views/AccountView.cs
+++ b/NickvisionMoney.GNOME/Views/AccountView.cs
@@ -834,7 +834,6 @@ public partial class AccountView : Adw.BreakpointBin
             {
                 path += ".pdf";
             }
-
             var dialog = Adw.MessageDialog.New(_parentWindow, _("Add Password To PDF?"), _("Would you like to password-protect the PDF file?\n\nIf the password is lost, the PDF will be inaccessible."));
             dialog.SetIconName(_controller.AppInfo.ID);
             dialog.AddResponse("no", _("No"));
@@ -856,7 +855,6 @@ public partial class AccountView : Adw.BreakpointBin
                 {
                     _controller.ExportToPDF(path ?? "", exportMode, null);
                 }
-
                 dialog.Destroy();
             };
             dialog.Present();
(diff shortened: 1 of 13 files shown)


JoseBritto:
Now this is just to make debugging for #762 a tiny bit easier. Will go  through and complete the PR by tuesday :)

nlogozzo:
Might be best to Ctrl+Shift+F for `catch` and add and print the exception where needed...

JoseBritto:
> Might be best to Ctrl+Shift+F for catch and add and print the exception where needed...
Alright! Also probably better to use Console.Error. I forgot that earlier

nlogozzo:
Console.WriteLine is probably fine...stdput and stderr point to the same terminal unless the programmer usually redirects it to a separate file which we don't really in this case...

JoseBritto:
> Console.WriteLine is probably fine...stdput and stderr point to the same terminal unless the programmer usually redirects it to a separate file which we don't really in this case...

It is actually two different streams but most terminals won't differentiate by default. Its standard thing for most programs to print use the correct output stream. Also users can manually redirect errors as well to a file or whatever. Since Denaro isn't a command line app, it probably doesn't really matter. Since I'm gonna have to go through everything anyway, might as well fix the few logs already being printed out.

JoseBritto:
I'll complete this in the afternoon. Have been a bit busy

JoseBritto:
Added Console.Error.WriteLine to most catch blocks that made sense to me. I ignored the ones that executed sql queries and such.

NickvisionMoney.GNOME/Views/AccountView.cs
```diff
@@ -828,6 +834,7 @@ private async Task ExportToPDFAsync(ExportMode exportMode)
             {
                 path += ".pdf";
             }
+
```
nlogozzo, line 837: Can you remove these extra white spaces?

NickvisionMoney.GNOME/Views/AccountView.cs
```diff
@@ -849,11 +856,15 @@ private async Task ExportToPDFAsync(ExportMode exportMode)
                 {
                     _controller.ExportToPDF(path ?? "", exportMode, null);
                 }
+
```
nlogozzo, line 859: Here too

NickvisionMoney.GNOME/Views/NewAccountDialog.cs
```diff
@@ -269,14 +269,18 @@ private async void SelectFolder(object? sender, EventArgs e)
             var folder = Gio.FileHelper.NewForPath(_controller.Folder);
             folderDialog.SetInitialFolder(folder);
         }
+
```
nlogozzo, line 272: Here too

nlogozzo (changes requested):

JoseBritto:
Idk why rider keeps adding these in

nlogozzo (approved):
Thanks!

nlogozzo:
https://github.com/NickvisionApps/Denaro/releases/tag/2024.1.0

PR #781, opened 2024-02-03
Add group.h and group.cpp

JoseBritto:


libdenaro/include/models/group.h
```diff
@@ -0,0 +1,107 @@
+#ifndef GROUP_H
+#define GROUP_H
+
+#include "color.h"
+#include <string>
```
nlogozzo, line 5: I like to order it with system includes first then local, so string first then color.h

libdenaro/include/models/group.h
```diff
@@ -0,0 +1,107 @@
...
+        bool operator>(const Group& other) const;
+        
+    private:
+        unsigned int m_id;
+        std::string m_name;
+        std::string m_description;
+        double m_income;
+        double m_expense;
+    }
+
```
nlogozzo, line 103: Remove extra line

libdenaro/include/models/group.h
```diff
@@ -0,0 +1,107 @@
...
+    private:
+        unsigned int m_id;
+        std::string m_name;
+        std::string m_description;
+        double m_income;
+        double m_expense;
+    }
+    
+}
+
```
nlogozzo, line 105: Remove extra line

nlogozzo (commented):
Pretty solid job 👌

libdenaro/include/models/group.h
JoseBritto: ok. will do

JoseBritto:
Done. Builds successfully as well

libdenaro/include/models/group.h
```diff
@@ -0,0 +1,105 @@
...
+         */
+        bool operator>(const Group& other) const;
+        
+    private:
+        unsigned int m_id;
+        std::string m_name;
+        std::string m_description;
+        double m_income;
+        double m_expense;
+        Color m_color;
```
JoseBritto, line 102: I have a doubt if this is correct or should it be Color&

nlogozzo: This is correct. The Color needs to be stored as the actual object, not a reference to an object, so no &.

The & is used in get and set method to avoid copying the color and just have a reference to use the one from the object itself.

nlogozzo (approved):

PR #784, opened 2024-02-12
C# --> C++

nlogozzo:


janopae:
Just out of curiousity: Why are switching to C++, and why does this enable Windows support? (I thought that C# had pretty good Windows support actually)

nlogozzo:
@janopae 

> why does this enable Windows support?

The switch to C++ itself does not enable Windows support. As you said yourself, C# (as well as C++) have first-class Windows support with the WindowsAppSDK, meaning either one could be used to make the same Windows app.

> Why are switching to C++

With that being said, we are switching to C++ for a couple of reasons:
1. In the Linux world, its much easier to build the C++ apps as opposed to C# apps in terms of dependencies and for user setup. Some users wish to stray away from Microsoft technologies, and C# is all Microsoft, whereas C++ is more decentralized.
2. Speed. The performance boost I've seen already with loading the same account with C++ vs C# is remarkable.

Plus, personally, I've always liked C++ more as a language than C# 😆

PR #799, opened 2024-03-16
Starting Dashboard Work

JoseBritto:


JoseBritto:
I tried to squash the commits into one but ended up making more 🤦

nlogozzo:
> I tried to squash the commits into one but ended up making more 🤦

Don't worry...I will squah and merge this whole pr once it's done

libdenaro/src/models/dashboardgroup.cpp
```diff
@@ -0,0 +1,25 @@
...
+    const std::unordered_map<Nickvision::Money::Shared::Models::Currency, std::pair<double, std::string>>& DashboardGroup::getData() const
+    {
+        return m_data;
+    }
+
+    void DashboardGroup::addAmount(const Nickvision::Money::Shared::Models::Currency& currency, double amount, const std::string& accountName)
+    {
+        if(!m_data.contains(currency))
+        {
+            m_data.emplace(currency, std::make_pair(amount, currency.getSymbol() + " " + std::to_string(amount) + " from " + accountName));
```
nlogozzo, line 19: The account string needs to be translated:

Add the following includes to the beginning of the file:
```c
#include <format>
#include <libnick/localization/gettext.h>
#include "helpers/currencyhelpers.h"
```

then replace the string in `make_pair` with `std::vformat(_("{} from {}"), std::make_format_args(CurrencyHelpers::toAmountString(amount, currency), accountName));`

You also need to add this file path to `/resources/po/POTFILES` that way it gets picked up in the po files.

libdenaro/src/models/dashboardgroup.cpp
```diff
@@ -0,0 +1,25 @@
...
+            return;
+        }
+        m_data.at(currency).first += amount;
+        m_data.at(currency).second = "\n" + currency.getSymbol() + " " + std::to_string(m_data.at(currency).first) + " from " + accountName;
```
nlogozzo, line 23: Same here...this line should be `m_data.at(currency).second += "\n" + std::vformat(_("{} from {}"), std::make_format_args(CurrencyHelpers::toAmountString(amount, currency), accountName));`

nlogozzo (changes requested):

nlogozzo:
@JoseBritto those new 3 includes should go in the cpp file, not the h file since that's where they are used

nlogozzo:
And the "\n" shouldn't be in the vformat otherwise it creates 2 translated strings instead of one...should be the way I wrote it

JoseBritto:
Messed up git again. [fd0456c](https://github.com/NickvisionApps/Denaro/pull/799/commits/fd0456c537e039e7ab7f5e2ec480f919b0ddf2ea) is the correct one

PR #806, opened 2024-05-23
Handle segfaults on startup

JoseBritto:
Things handled:
- [x] Segfault when missing gresource file
- [x] Segfault when missing *.ui files
- [x] Segfault when locale not setup on system ~*(im not entirely sure how to handle this, working on it atm)*~

Let me know if there are other things I missed, but these were mostly the problems I faced and having no logs made it harder to debug.


nlogozzo:
> Segfault when locale not setup on system (im not entirely sure how to handle this, working on it atm)

You might have to take a look at `libnick` repository, as that is where all the localization happens. But it already handles if no local exists and defaults to en-us.

JoseBritto:
> > Segfault when locale not setup on system (im not entirely sure how to handle this, working on it atm)
> 
> You might have to take a look at `libnick` repository, as that is where all the localization happens. But it already handles if no local exists and defaults to en-us.

Oh. For me if I unset the locale, Denaro crashes. Basically std::locale("") throws runtime error causing a crash. How would I handle this?

JoseBritto:
In denaro there are two places that directly call std::locale. In the CurrencyHelpers::getSystemCurrency() method and the MainWindowController::getDebugInformation() method, I tried setting std::locale::global(std::locale("C")) but that doesn't seem to affect anything. I'm kinda lost here.  Gtk by default falls back to this and produces an error:

```
(org.nickvision.money.gnome:27160): Gtk-WARNING **: 23:32:48.436: Locale not supported by C library.
	Using the fallback 'C' locale.
```

Is the correct way to have a small helper method/class that proxies all the calls to get the current locale and handle it there? Or is it better to handle the exception where ever we use std::locale directly?

JoseBritto:
> In denaro there are two places that directly call std::locale. In the CurrencyHelpers::getSystemCurrency() method and the MainWindowController::getDebugInformation() method, I tried setting std::locale::global(std::locale("C")) but that doesn't seem to affect anything. I'm kinda lost here. Gtk by default falls back to this and produces an error:
> 
> ```
> (org.nickvision.money.gnome:27160): Gtk-WARNING **: 23:32:48.436: Locale not supported by C library.
> 	Using the fallback 'C' locale.
> ```
> 
> Is the correct way to have a small helper method/class that proxies all the calls to get the current locale and handle it there? Or is it better to handle the exception where ever we use std::locale directly?

Handled both separately.

JoseBritto:
Please let me know if I used the & and * or none at all for variables properly. I still get confused sometimes 😅 
Everything worked when I tested.

org.nickvision.money.gnome/src/application.cpp
```diff
@@ -19,7 +19,13 @@ namespace Nickvision::Money::GNOME
         }
         m_controller->getAppInfo().setChangelog("- Initial Release");
         std::filesystem::path resources{ Aura::getActive().getExecutableDirectory() / (m_controller->getAppInfo().getId() + ".gresource") };
-        g_resources_register(g_resource_load(resources.string().c_str(), nullptr));
+        GError* resourceLoadError{nullptr};
```
nlogozzo, line 22: Styling:
should be `{ nullptr }` not `{nullptr}`

org.nickvision.money.gnome/src/application.cpp
```diff
@@ -19,7 +19,13 @@ namespace Nickvision::Money::GNOME
...
+        GResource* resource = g_resource_load(resources.string().c_str(), &resourceLoadError);
```
nlogozzo, line 23: `GResource* resource{ g_resource_load(resources.string().c_str(), &resourceLoadError) };`

org.nickvision.money.gnome/src/application.cpp
```diff
@@ -19,7 +19,13 @@ namespace Nickvision::Money::GNOME
...
+        if(resourceLoadError != nullptr)
+        {
+            throw std::runtime_error(resourceLoadError->message); // Panic
```
nlogozzo, line 26: you don't need the `// Panic` comment here

org.nickvision.money.gnome/src/helpers/builder.cpp
```diff
@@ -13,6 +13,7 @@ namespace Nickvision::Money::GNOME
         std::filesystem::path path{ Aura::getActive().getExecutableDirectory() / "ui" / (blueprint + ".ui") };
         if(!std::filesystem::exists(path))
         {
+            fprintf(stderr, "UI file not found at path: %s", path.c_str());
```
nlogozzo, line 16: you can make this C++ instead of C:

```cpp
std::err << "UI file not found at path: " << path << std::endl;
```

Just make sure to add `#include <iostream>` after `#include <filesystem>`

org.nickvision.money.gnome/src/views/mainwindow.cpp
```diff
@@ -36,6 +36,10 @@ namespace Nickvision::Money::GNOME::Views
         m_currencyConverterPage{ GTK_WINDOW(m_window) }
     {
         //Setup Window
+        if(m_window == nullptr)
```
nlogozzo, line 39: There's no need for this if statement since the `fromBlueprint` function will throw an error if the blueprint files didn't get compiled.

libdenaro/src/helpers/currencyhelpers.cpp
```diff
@@ -44,24 +44,36 @@ namespace Nickvision::Money::Shared
...
-            systemCurrency->setDecimalSeparator(std::use_facet<std::moneypunct<char>>(std::locale("")).decimal_point());
-            systemCurrency->setGroupSeparator(std::use_facet<std::moneypunct<char>>(std::locale("")).thousands_sep());
+            std::locale locale;
+            try
+            {
+                locale = std::locale("");
+            }
+            catch (...)
+            {
+                fprintf(stderr, "Unknown locale on system. Defaulting to USD as the currency");
```
nlogozzo, line 54: You can make this C++ instead of C:
```cpp
std::err << "Unknown system locale. Reverting to USD currency." << std::endl;
```

Just be sure to add `#include <iostream>` after `#include <iomanip>`.

nlogozzo (changes requested):

org.nickvision.money.gnome/src/application.cpp
```diff
@@ -19,7 +19,13 @@ namespace Nickvision::Money::GNOME
         }
         m_controller->getAppInfo().setChangelog("- Initial Release");
         std::filesystem::path resources{ Aura::getActive().getExecutableDirectory() / (m_controller->getAppInfo().getId() + ".gresource") };
-        g_resources_register(g_resource_load(resources.string().c_str(), nullptr));
+        GError* resourceLoadError{nullptr};
+        GResource* resource = g_resource_load(resources.string().c_str(), &resourceLoadError);
+        if(resourceLoadError != nullptr)
```
nlogozzo, line 24: You can simplify this to `if(resourceLoadError)` which will hit true if `resourceLoadError` is not null.

nlogozzo (changes requested):

JoseBritto:
All done

nlogozzo (changes requested):

libdenaro/src/helpers/currencyhelpers.cpp
```diff
@@ -1,5 +1,6 @@
 #include "helpers/currencyhelpers.h"
 #include <iomanip>
+#include <iostream>
```
nlogozzo, line 3: replace this with `#include <libnick/app/aura.h>`

but place it above `#include <libnick/helpers/stringhelpers.h>`

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -206,22 +206,33 @@ namespace Nickvision::Money::Shared::Controllers
...
         {
+            Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "Checking for updates...");
             std::thread worker{ [&]()
             {
                 Version latest{ m_updater->fetchCurrentStableVersion() };
                 if (!latest.empty())
                 {
                     if (latest > Aura::getActive().getAppInfo().getVersion())
                     {
+                        Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "Updates found. Latest version is " + latest.toString());
```
JoseBritto, line 225: everything except this line is the same as you said

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -197,22 +206,33 @@ namespace Nickvision::Money::Shared::Controllers
...
                         m_notificationSent.invoke({ _("New update available"), NotificationSeverity::Success, "update" });
                     }
+                    else
+                    {
+                        Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "No updates found");
```
nlogozzo, line 230: Missing `.` at end

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -197,22 +206,33 @@ namespace Nickvision::Money::Shared::Controllers
...
+                    }
+                }
+                else
+                {
+                    Aura::getActive().getLogger().log(Logging::LogLevel::Warning, "Unable to fetch latest app version");
```
nlogozzo, line 235: Missing `.`. at end

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -296,13 +333,15 @@ namespace Nickvision::Money::Shared::Controllers
...
         {
+            Aura::getActive().getLogger().log(Logging::LogLevel::Error, std::string (e.what()) + " : " + newAccountDialogController->getFilePath().string());
             m_notificationSent.invoke({ e.what(), NotificationSeverity::Error });
         }
         if(controller)
         {
             m_accountViewControllers[newAccountDialogController->getFilePath()] = controller;
             config.addRecentAccount(controller->toRecentAccount());
             config.save();
+            Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "Config saved");
```
nlogozzo, line 344: Missing `.` at end

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -338,6 +378,7 @@ namespace Nickvision::Money::Shared::Controllers
                 Configuration& config{ Aura::getActive().getConfig<Configuration>("config") };
                 config.addRecentAccount(controller->toRecentAccount());
                 config.save();
+                Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "Config saved");
```
nlogozzo, line 381: Missing `.` at end

libdenaro/src/controllers/mainwindowcontroller.cpp
```diff
@@ -349,6 +390,7 @@ namespace Nickvision::Money::Shared::Controllers
         Configuration& config{ Aura::getActive().getConfig<Configuration>("config") };
         config.removeRecentAccount(account);
         config.save();
+        Aura::getActive().getLogger().log(Logging::LogLevel::Debug, "Config saved");
```
nlogozzo, line 393: Missing `.` at end

nlogozzo (changes requested):

JoseBritto:
[e777bf7](https://github.com/NickvisionApps/Denaro/pull/806/commits/e777bf7bf002d920f8d3c875452e87bca41afa27) Added some more logs when opening and removing accounts. Let me know if I went a little overboard there.

nlogozzo (approved):

nlogozzo:
> [e777bf7](https://github.com/NickvisionApps/Denaro/pull/806/commits/e777bf7bf002d920f8d3c875452e87bca41afa27) Added some more logs when opening and removing accounts. Let me know if I went a little overboard there.

All is fine. Just only removed the `Account removed` one that had the account name, because i'd like to keep as little personal information as possible from the logs (at most file paths).

